Issue #165, opened 2026-04-17
Portable bench: long-horizon scenario runner with per-system telemetry and plots


hardbyte:
## Summary

Add a **long-horizon scenario runner** to `benchmarks/portable/` that drives multi-hour workloads against each queue system, collects Postgres-side and adapter-side telemetry at fixed cadence, and produces publication-quality cross-system plots. Idle-in-tx is the first interesting scenario to build with it, but the runner is deliberately general — other long-running failure modes (sustained high load, overlapping active readers, checkpoint storms, long steady-state soak) compose from the same phase primitives.

This issue covers the **harness, scenarios, metrics, and plots** only — the new adapters (pgmq, pg-boss, PgQ) ship in #166 which depends on the contracts defined here.

## Motivation

The current portable bench (`benchmarks/portable/run.py`) measures short steady-state throughput. That hides every interesting failure mode that develops over minutes-to-hours: MVCC bloat under a held xmin horizon, index fragmentation after prolonged churn, checkpoint-triggered latency spikes, soak drift. To compare awa against peer systems on these axes — and to generate the plots that accompany ADR discussions — we need a harness that:

1. Runs multi-hour scenarios with explicit phases (warmup, clean, pressure, recovery, …).
2. Applies pressure in well-defined ways per phase type (idle-in-tx session, active-readers, sustained high load).
3. Collects PG-side + adapter-side telemetry on a shared timebase so cross-system plots are comparable.
4. Produces self-explanatory plots suitable for a blog post or ADR without re-processing.

The first concrete failure mode to measure is idle-in-tx: any session in `idle in transaction` pins the xmin horizon, autovacuum stalls on every table, and UPDATE/DELETE-based queue designs accumulate dead tuples. The rotating-event-table pattern (PgQ, 2007) sidesteps this because `TRUNCATE` bypasses MVCC. But the harness itself is agnostic — once built, sustained-high-load, active-readers, and soak scenarios follow trivially.

This issue is **observation only** — no architectural changes to awa. That lands in a follow-up ADR informed by the measurements.

### Relationship to existing awa MVCC benchmarks

awa already has a dedicated Rust-test MVCC benchmark lane:

[...]

PR #170, opened 2026-04-18
Portable bench: long-horizon scenario runner with per-system telemetry and plots

hardbyte:
## Summary
- add the long-horizon portable benchmark harness with per-system adapter manifests, phase hooks, telemetry collection, summaries, and publication-quality plots
- wire the new harness into docs and CI, then harden adapter shutdown, native binary resolution, Postgres readiness, and producer control so high-load and depth-target modes work end-to-end
- add zipped HTML reporting for the full portable suite and validate the branch with a clean consolidated rerun while leaving generated results out of git

## Validation
- `uv run --project benchmarks/portable --extra dev python -m pytest benchmarks/portable/tests/test_harness_smoke.py`
- short long-horizon verification runs across the default systems plus focused `high-load` and `depth-target` checks
- `uv run python benchmarks/portable/full_suite.py`

## Summary by CodeRabbit

* **New Features**
  * Long‑horizon portable benchmark runner with phase DSL, adapter contract, aligned sampling, automated plots and HTML report bundle; multiple new adapters and adapter contributor guide; Python package for the harness.

* **Improvements**
  * Improved container entrypoints, native binary discovery, portable Postgres config, and per-run manifest/summary generation.

* **Tests**
  * New CI smoke job and nightly reduced long‑horizon workflow; added harness smoke tests for parsing, CSV/schema, summary, and plotting.


hardbyte:
## Overall

Approving with a short list of things worth addressing before merge. This is a large, well-structured PR that matches issue #165's spec closely. Concerns are mostly minor; the one real blocker I found is a silent-failure path in `enter_idle_in_tx` that would invalidate the headline measurement.

CodeRabbit and Codex between them have already flagged most of the substantive issues (19 inline comments). I'm not going to repeat those here. Prioritisation and new findings follow.

## New (not already flagged)

1. **`enter_idle_in_tx` can silently no-op the idle phase** (`bench_harness/hooks.py:31-49`). Blocker. If the holder thread's `psycopg.connect` or `SELECT txid_current()` raises, the thread dies without signaling back. `enter_idle_in_tx` returns normally and the phase runs with no held transaction. The idle-in-tx measurement is the whole point of the harness on day one. A worker DB restart between `preflight_database` and phase start, or a transient connection error, would produce results that look clean but aren't. Fix: wait briefly (few hundred ms) for `holder["xid"]` or a failure flag to be set, raise if neither.

   ```python
   def enter_idle_in_tx(runtime: PhaseRuntime) -> None:
       ready = threading.Event()
       error: dict[str, Exception] = {}
       ...
       def _hold() -> None:
           try:
               with psycopg.connect(..., autocommit=False) as conn:
                   with conn.cursor() as cur:
                       cur.execute("SELECT txid_current()")
                       holder["xid"] = cur.fetchone()[0]
                   ready.set()
                   stop.wait()
           except Exception as exc:
               error["exc"] = exc
               ready.set()
       ...
       if not ready.wait(timeout=5.0):
           raise RuntimeError("idle-in-tx holder did not start in 5s")
       if "exc" in error:
           raise RuntimeError(f"idle-in-tx holder failed: {error['exc']}") from error["exc"]
   ```

2. **`phase_boundary_snapshot` doesn't align timestamps** (`bench_harness/metrics.py:131-192`). Related to Codex's P2 but more specific. The continuous `_poll_once` uses `_tick_elapsed_s` so every emit in one tick shares one elapsed_s. `phase_boundary_snapshot` doesn't set it, so each `_emit` in the snapshot gets a fresh `time.time() - bench_start`. Ordering is correct but downstream aggregations that key on exact `elapsed_s` equality will see phase-boundary rows scattered across microsecond-distinct timestamps instead of coalescing. Fix: wrap the snapshot body with the same try/finally that sets `_tick_elapsed_s`/`_tick_sampled_at` once.

3. **`enter_active_readers` default scan is anemic** (`bench_harness/hooks.py:88-91`). Default `ACTIVE_READER_SQL` is `SELECT count(*) FROM pg_stat_user_tables` — a catalog scan. The PlanetScale failure mode is about analytics readers against queue DATA, not the system catalog. Catalog-only scans still hold a snapshot so the xmin pin works, but the reader does no real I/O on the high-churn tables, which changes the concurrency picture (less lock contention, no heap buffer pressure). Suggest defaulting to `SELECT count(*) FROM <first event table from manifest>` and fall back to the catalog scan only if that's empty.

4. **`snapshot_xmin` precision loss in `_CLUSTER_SQL`** (`bench_harness/metrics.py:62`). `pg_snapshot_xmin(pg_current_snapshot())::text::double precision` round-trips an xid8 (uint64) through text to float64. Float64 mantissa is 53 bits so xid values above 2^53 lose precision. Won't matter for any plausible run, but worth a comment near the cast so a future maintainer doesn't convert the metric to something that assumes integer semantics.

5. **Codex P1 "Normalize adapter system" (`comment 3104160710`) is already fixed.** Lines 422-426 of `metrics.py` explicitly override with `expected_system` and the inline comment even notes "Different adapter binaries (e.g. awa-docker vs native awa) can both emit system='awa'". Can resolve that thread without a code change.

## Prioritisation of existing CodeRabbit/Codex comments

Blockers (land before merge):

- `full_suite.py:424` ZeroDivisionError when all chart values are zero (CodeRabbit Critical). Easy fix, real runtime crash.
- `orchestrator.py:319` descriptor timeout/drift doesn't kill the adapter process (CodeRabbit Major). Leaves a running subprocess on the host if drift fails.
- `metrics.py:199` silent DB-connect failure in daemon (CodeRabbit Major). Same class as my #1 above; at minimum log to stderr.
- `metrics.py:219` catch-up replay after slow poll (CodeRabbit Major). `next_tick += period` in a loop produces back-to-back polls instead of skipping to the next boundary. Fix: `while next_tick <= now: next_tick += period`.
- `river-bench/main.go:491` `SAMPLE_EVERY_S` accepts 0/negative (CodeRabbit Major). `time.NewTicker(0)` panics.

Should fix before merge:

- `plots.py:431` faceted plot double-init of axes (CodeRabbit Major). Duplicated phase bands + wrong legend labels.
- `metrics.py:429` malformed `window_s` raises instead of returning None (CodeRabbit Major). A single bad adapter line kills sample ingestion.
- `river-bench/main.go:617` uncancellable initial alignment sleep (CodeRabbit Major). Up to `SAMPLE_EVERY_S` of unresponsive shutdown.

Minor (queue for a follow-up cleanup, not blocking):

- URL parsing (`river-bench` L503, `awa-python-bench` L318) — `db_name` captures `?query-params`.
- `awa-python-bench` unused `padding`.
- `nightly-bench-harness.yml` scenario input is misleading.
- `README.md` scenario duration estimates.
- `CONTRIBUTING_ADAPTERS.md` code fence language tag.
- EN DASH character in `full_suite.py`.
- `awa-python-bench` type hint deque-before-import.

## Architectural notes (positive)

Not actionable — just calling out what I think is well-chosen and would be a shame to lose in future refactors:

- **Static adapter.json as authoritative for preflight, runtime descriptor cross-checked as superset.** Breaks the bootstrap circularity a runtime-only contract would have; catches drift. Good.
- **Clock-aligned sampling via `_next_aligned_tick` in both metrics daemon and Rust adapter sampler.** Cross-system plots line up without post-hoc interpolation.
- **`_tick_elapsed_s` + `_tick_sampled_at` sharing per-poll timestamp** is the right mechanism; just missing from `phase_boundary_snapshot` (above).
- **Family + variant-linestyle palette for same-system SDKs** (`awa` family covering native / Docker / Python with colour + linestyle). Clean cross-system plotting when multiple flavours of the same implementation are on the chart.
- **pinned `postgres:17.2-alpine` + committed `postgres.conf` with explicit defaults.** Prevents silent drift from upstream PG changes.
- **Builder/launcher split in adapters.py** keeps build-once-launch-many tidy, and `_cargo_target_dir` shelling out to `cargo metadata` correctly handles workspace + `CARGO_TARGET_DIR` overrides.
- **`phases.py` runtime hooks imported lazily** so the fixture-only CI smoke job doesn't need psycopg. That pays off for keeping the blocking smoke job under 15s.

## CI

All 13 checks on the latest run were green or in-progress at review time. `bench-harness-smoke` passes in ~10s. No DB-bound test coverage for the harness beyond fixture-based smoke, which is the right call — the real validation is the nightly cross-system run.

## Suggested merge strategy

Fix the five blockers and my #1 before merge. The shoulds I'd take, the minors can live as a single follow-up PR. Once blockers are in, the branch is ready.

Issue #174, opened 2026-04-18
Unify chaos and long-horizon runners: one phase-driven driver, one adapter contract


hardbyte:
## Summary

The portable bench currently runs two parallel drivers: `long_horizon.py` (phased workloads with telemetry sampling) and `chaos.py` (short scenarios that disrupt the adapter or Postgres). They share their infrastructure (adapter registry, manifest, launcher, PG lifecycle, CLI validation, output writing) but duplicate the scaffolding and accrue per-adapter hardcoding in the chaos side.

One phase-driven driver can replace both. The difference between the two systems is the kind of phases that run, not the system that runs them.

## Current state

**Infrastructure already shared by design** (landed in #170):

- Adapter registry (`ADAPTERS`, `AdapterManifest`).
- Static per-adapter manifest (`adapter.json` with `db_name`, `event_tables`, `extensions`).
- Launcher/builder split with docker vs native distinction.
- PG container lifecycle.
- Pydantic-validated CLI config.
- raw.csv + summary.json output pipeline.

**What `chaos.py` duplicates** (2010 lines, ~142 per-system refs):

- A second `SYSTEMS` dict of per-adapter SQL templates and state-column names, parallel to `adapter.json`.
- Six per-system `start_<system>_worker` functions with inline docker argv.
- Per-scenario `if system == "awa": ... elif ...` ladders inside every chaos scenario function.
- Its own output format (scenario-level JSON, not time-series).

**What chaos actually does that long-horizon doesn't:**

- Destructive operations on the adapter process (SIGKILL, then relaunch and verify recovery).
- Destructive operations on Postgres (container restart, `pg_terminate_backend`, pool exhaustion).
- Pass/fail scenario framing rather than continuous measurement.

All three of those are expressible as phase types in the existing DSL.

## Desired outcome

One driver. One CLI entry point. One adapter contract. Chaos-like behaviour (destructive events) becomes a set of additional phase types in `bench_harness/hooks.py`, alongside the existing `idle-in-tx`, `active-readers`, and `high-load`.

Shape, not prescription:

- Existing phases: `warmup`, `clean`, `idle-in-tx`, `recovery`, `active-readers`, `high-load`.
- Chaos-class phases (new): destructive events against the worker or Postgres; the orchestrator owns the subprocess handle and can relaunch as part of the phase lifecycle.
- Existing chaos "scenarios" become named phase sequences, the same way long-horizon scenarios already are.
[...]

Issue #171, opened 2026-04-18
Portable bench report: interactive HTML (expandable plots, filterable tables)


hardbyte:
Follow-up from #170 (portable bench long-horizon harness).

The current HTML report bundle (`benchmarks/portable/full_suite.py` → `write_html_report`) is a static page with inline SVG charts and HTML tables. It would be much nicer to make it interactive:

## Wishlist

- **Click-to-expand charts** — open any SVG in a modal/lightbox at full size, ideally with hover tooltips per bar showing the exact value.
- **Filterable / sortable tables** — at least the cross-system benchmark and cross-system chaos tables should be sortable by column and filterable by system / scenario / status.
- **Status badge filter** — quick toggle to show only failing/partial chaos rows.
- **Per-system jump menu** — sticky nav linking to each `#system-…` anchor.
- **Optional dark mode** — the bundle is often viewed in terminals/IDEs.

## Implementation notes

- Keep the bundle self-contained (single zip, no external CDNs) so it works offline.
- Likely a small vanilla-JS file in `assets/` plus minor markup changes; avoid heavy frameworks.
- For chart interactivity, consider switching the inline SVG generation to embed data attributes the JS can read for tooltips, rather than swapping in a charting library.

## Out of scope

- Re-running benchmarks. This is purely a reporting/UX change.

hardbyte:
Closed: interactive HTML report has been implemented in the [postgresql-job-queue-benchmarking](https://github.com/hardbyte/postgresql-job-queue-benchmarking) repo, which owns the cross-system reporting going forward.

Issue #167, opened 2026-04-17
Chaos tests rely on stdout parsing + timing; redesign to DB-state observation


hardbyte:
## Summary

The chaos tests in \`awa-python/tests/test_chaos_recovery.py\` synchronize via stdout-parsing of worker log lines (\"START\", \"WAITING\", \"COMPLETE\" with PIDs and job IDs). This is a scaffolding-heavy design that couples correctness assertions to (a) process scheduling, (b) Rust tracing-subscriber buffering on non-TTY stderr, (c) Linux pipe capacity (64 KiB default), (d) the test process consuming fast enough. None of those prove the behaviour under test — they're observability scaffolding.

Symptoms we've seen during PR #163 CI:

- \`test_callback_timeout_is_rescued_and_retried\`: worker logs COMPLETE for attempt=2, DB row stuck at \`state=running, attempt=2, finalized_at=<3s into test>\` for 30s. Partially explained by pipe-buffer saturation: once the test matched its expected line it stopped draining the pipe, Rust tracing at ~50 lines/sec filled the 64 KiB buffer in ~9s, the next write blocked, and any tokio worker thread holding that write stalled — including potentially the completion batcher.
- \`test_worker_sigkill_job_is_rescued_and_completed\`: same symptom shape, different test. Pipe-buffer fix in PR #163 may or may not have been sufficient — diagnostic-only run went green but we can't prove causality from CI alone.

## Proposal

Redesign the three chaos tests to observe via the DB, with stdout retained only for the \"process is alive\" READY signal.

| Test | What it's proving | Observable via DB |
|---|---|---|
| \`test_worker_sigkill\` | SIGKILL'd worker's job is rescued and completed by a replacement | \`state=running AND attempt=1 AND heartbeat_at recent\` → kill → \`state=completed AND attempt=2\` |
| \`test_worker_hang\` | Hung handler cancelled via deadline rescue and retried | DB transitions + one assertion that handler observed cancel signal (DB-backed, not stdout) |
| \`test_callback_timeout\` | Callback expires → rescue → retry → complete | \`running → waiting_external → retryable → running → completed\` |

## Implementation notes

- Keep \`_wait_for_line(READY)\` for the single \"runtime is up and leader\" bootstrap check. Everything else switches to \`_wait_for_db(predicate, timeout)\`.
- For \"handler observed cancel\" in the hang test: have the worker write to a side-channel (marker table, or a job \`metadata.cancel_observed = true\` update) rather than print to stdout.
[...]

Issue #166, opened 2026-04-17
Portable bench: pgmq, pg-boss, PgQ adapters


hardbyte:
## Summary

Add three adapters to `benchmarks/portable/` so the long-horizon harness (#165) can baseline awa against a broader field of Postgres-backed queue systems:

- **pgmq** — Tembo's Postgres extension
- **pg-boss** — Node.js, UPDATE-based on a single jobs table
- **PgQ** — Skype's original rotating-event-table queue, for the MVCC-bloat-immune reference line

Depends on #165 for the adapter contract (long-horizon mode, JSONL telemetry, declared event tables, declared extensions). The full contract is documented in `benchmarks/portable/CONTRIBUTING_ADAPTERS.md` once #165 lands — each of the three adapters here doubles as a reference implementation of that contract for future contributors (pgmq = PG-extension style, pg-boss = Node.js style, PgQ = rotating-event-tables style).

## Motivation

The idle-in-tx / long-horizon scenarios in #165 want to show how different Postgres queue architectures respond to a held xmin horizon. The existing adapter set (awa, awa-python, River, Procrastinate, Oban) covers a handful of UPDATE-based designs but not a wide enough spread to tell the architectural story. pgmq and pg-boss broaden the UPDATE camp; PgQ is the reference line that's immune because it rotates event tables and `TRUNCATE`s.

All three are mature public projects with their own official migration paths — this issue does **not** maintain custom SQL. Each adapter calls the system's own install/migration entry point.

---

## Adapter contract (from #165)

Each new adapter implements the same contract as the existing ones:

- Runs the system's **own** migrations / extension install (not harness-authored SQL).
- Declares its event tables via `event_tables()` so the metrics daemon knows what to poll.
- Declares required Postgres extensions via `extensions()` so the harness creates them on the system's database before `setup()`.
- Supports the long-horizon lifecycle: run steady-state producer + consumer until SIGTERM, emit JSONL samples every 10s to stdout with rolling 30s stats (P50/P95/P99 claim latency, enqueue/completion rates, queue depth).
- Accepts config via environment variables consistent with existing adapters (`DATABASE_URL`, `WORKER_COUNT`, `PRODUCER_RATE_CAP`, etc.).

Each adapter gets its own database in `init-databases.sql` and ships in its own directory under `benchmarks/portable/`. The PG container image is upgraded (or a custom `Dockerfile` is added for the PG service) to include any binary extensions the adapters need.

---

### pgmq (Postgres extension by Tembo)

[...]

PR #185, opened 2026-04-20
Portable benchmark unification and queue storage pressure fixes

hardbyte:
Advances #174, #171, #167, and #166.

## What this PR does

- folds the phase-driven portable benchmark work onto the queue-storage branch
- benchmarks Awa directly against the queue-storage engine in the long-horizon adapter
- adds producer, subscriber, and end-to-end latency metrics to the portable harness and report
- adds the interactive HTML report, denser sampling, timeline presets, outlier warnings, and hides warmup by default in presentation
- adds PgQue, pgmq, and pg-boss adapters to the event-delivery comparison harness
- fixes the real queue-storage pressure issue uncovered by the portable runs by moving completed-history rollups off the hot `queue_lanes` prune path and narrowing prune lock timeouts
- updates README, architecture, ADR-019, and the storage-model correspondence docs to match the current implementation

## Validation

- `cargo test -p awa --test queue_storage_runtime_test -- --nocapture`
- `cd benchmarks/portable && uv run --extra dev pytest tests/test_harness_smoke.py -q`
- longer pressure and comparison runs were exercised while developing this slice, including `awa` vs `pgque` event-delivery scenarios and transition/pressure checks on the queue-storage branch

## Notes

- the portable report now defaults to steady-state presentation: warmup remains in `raw.csv` for diagnosis but is hidden from `summary.json` and the HTML report by default
- the benchmark-driven queue-storage fix keeps terminal count rollups off the completion hot path; queue counts now rely on live terminal rows plus the cold `pruned_completed_count` rollup after successful prune


hardbyte:
Addressed the two inline review points on the prune-rollup change in `84de3bc`:

- legacy `queue_lanes.pruned_completed_count` values are now backfilled into `queue_terminal_rollups` during schema preparation, and `queue_counts()` temporarily reads both sources with a non-double-counting fallback so upgrades do not under-report completed totals
- `prune_oldest()` now updates the cold terminal rollup inside the same transaction as `TRUNCATE`, so completed-count preservation is atomic with prune

Added runtime coverage for the legacy backfill/fallback path in `test_queue_storage_queue_counts_reads_legacy_lane_rollups_and_backfills_them`.

Validation rerun: `cargo test -p awa --test queue_storage_runtime_test -- --nocapture`

PR #224, opened 2026-05-03
Restore capacity-release wake drains

### LINKED ISSUE #223: Throughput regression from capacity-wake suppression in #216
## Summary

A focused 128-worker bisect shows PR #216 (`ec7d9ae`, "Reduce empty capacity wake claims") is the load-bearing throughput regression between alpha.3 and current main.

| Commit | Label | Run 1 | Run 2 | Mean @ 128w |
| --- | --- | ---: | ---: | ---: |
| `d7748d0` | alpha.3 baseline | 4,973 | 4,766 | 4,870 |
| `8ebcbc5` | after #214, claimer cap | 5,316 | 5,012 | 5,164 |
| `d930e72` | after #215, striped claims | 4,676 | 4,997 | 4,837 |
| `ec7d9ae` | after #216, empty wake | 2,921 | 2,254 | 2,588 |
| `5a8db21` | after #220, current | 2,869 | 2,590 | 2,729 |

## Reading

- #214 and #215 are within run-to-run variance of the alpha.3 baseline.
- #216 is the cliff: `4,837 -> 2,588`, about a 47% drop from a single PR.
- #220 partially repaired one case, especially visible around 16 workers, but it did not address the bulk of the 64/128-worker regression.
- #222 did not validate the lock-window/deserialization hypothesis; it should not be treated as the perf recovery.

## Likely cause

The capacity-wake suppression introduced in #216 is globally too aggressive. Under high concurrency, completion-driven capacity wakes appear to be useful work-dispatch signals even after prior empty claims. Suppressing those wakes leaves workers under-refilled and shifts too much responsibility to the slower fallback poll path.

## Candidate fix

Revert or rewrite the capacity-release wake suppression. A narrower candidate is to keep the exponential fallback poll backoff from #216, but restore alpha.3-style behavior where every capacity release wake drains the dispatcher.

A probe branch exists at `brian/probe-unsuppress-capacity-wakes` for that shape.

hardbyte:
## Summary
- remove the #216 capacity-wake suppression state machine
- restore alpha.3-style behavior where every capacity release wake drains ready work
- keep the exponential fallback poll backoff from #216

Fixes #223.

## Why
A focused 128-worker bisect shows #216 is the throughput cliff:

| Commit | Label | Mean @ 128w |
| --- | --- | ---: |
| `d7748d0` | alpha.3 baseline | 4,870 |
| `8ebcbc5` | after #214 | 5,164 |
| `d930e72` | after #215 | 4,837 |
| `ec7d9ae` | after #216 | 2,588 |
| `5a8db21` | after #220/current | 2,729 |

#220 recovered the partial-claim case, but the high-concurrency regression remained. This patch isolates the remaining #216 behavior by keeping poll backoff while removing completion/capacity wake suppression.

## Validation
- `cargo test -p awa-worker dispatcher::tests:: -- --nocapture`
- `cargo clippy -p awa-worker --all-targets -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/awa-worker/src/dispatcher.rs
+++ b/awa-worker/src/dispatcher.rs
@@ -66,28 +66,6 @@ impl WakeReason {
     }
 }
 
-#[derive(Debug, Default)]
-struct CapacityWakeState {
-    recheck_on_capacity: bool,
-}
-
-impl CapacityWakeState {
-    fn should_drain_on_capacity(&self) -> bool {
-        self.recheck_on_capacity
-    }
-
-    fn mark_wake_deferred_for_capacity(&mut self) {
-        self.recheck_on_capacity = true;
-    }
-
-    fn record_claim_result(&mut self, claimed: usize) {
-        // A non-empty claim is enough evidence that capacity release may
-        // expose more ready work. Empty claims proved the opposite and should
-        // suppress completion-driven rechecks until another signal arrives.
-        self.recheck_on_capacity = claimed > 0;
-    }
-}
-
 #[derive(Debug, Clone)]
 struct PollBackoff {
     base: Duration,
@@ -342,7 +320,6 @@ pub struct Dispatcher {
     rate_limiter: Option<TokenBucket>,
     storage: RuntimeStorage,
     capacity_wake: Arc<Notify>,
-    capacity_wake_state: CapacityWakeState,
     poll_backoff: PollBackoff,
 }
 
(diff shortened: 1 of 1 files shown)


hardbyte:
Closing as superseded by #225. #224 was useful as the capacity-wake A/B slice, but #225 includes that restoration plus the fixed-polling change that closed #223.

PR #235, opened 2026-05-06
Perf bundle: multi-queue notify fanout + completion CTE merge

hardbyte:
## Summary

Queue-storage performance bundle containing the remaining storage-path changes after `awa-metrics` landed separately in #232. Against current `main`, this PR only changes `awa-model/src/queue_storage.rs`.

Included changes:

- **Fold multi-queue NOTIFY fanout into one statement.** Ready enqueue paths now collect destination queues and call `notify_queues_tx`, which deduplicates logical channels and emits `pg_notify` through a single `unnest($1::text[])` query. Single-queue behavior remains one NOTIFY for one logical queue.
- **Merge `attempt_state` cleanup into runtime lease deletion.** `complete_runtime_batch` now deletes matching `attempt_state` rows inside the lease-delete CTE, removing one database round trip per completion batch while preserving the same transaction boundary and lease-delete `RETURNING` shape.

## Validation

- CI green on this bundle
- CodeRabbit review completed
- Constituent PRs #233 and #234 were CI green
- `git diff --check origin/main..pr/235` clean
- Existing queue-storage runtime coverage exercises completion cleanup, receipt fallback, stale-result guards, and striped runtime paths

## Notes

No end-user documentation update is needed: these changes do not add or change public configuration, API, schema, or operational semantics. The expected benefit is fewer database round trips in multi-queue enqueue batches and completion-heavy workloads.

### DIFF
--- a/awa-model/src/queue_storage.rs
+++ b/awa-model/src/queue_storage.rs
@@ -4773,17 +4773,8 @@ impl QueueStorage {
         let mut tx = pool.begin().await.map_err(map_sqlx_error)?;
         let total_rows = self.insert_ready_rows_tx(&mut tx, rows.clone()).await?;
 
-        let queues_to_notify: BTreeSet<String> = rows
-            .iter()
-            .map(|row| self.logical_queue_name(&row.queue).to_string())
-            .collect();
-        for queue in queues_to_notify {
-            sqlx::query("SELECT pg_notify($1, '')")
-                .bind(format!("awa:{queue}"))
-                .execute(tx.as_mut())
-                .await
-                .map_err(map_sqlx_error)?;
-        }
+        let queues_to_notify: Vec<String> = rows.iter().map(|row| row.queue.clone()).collect();
+        self.notify_queues_tx(&mut tx, queues_to_notify).await?;
 
         tx.commit().await.map_err(map_sqlx_error)?;
         Ok(total_rows)
@@ -4892,10 +4883,8 @@ impl QueueStorage {
             }
         }
 
-        let queues_to_notify: BTreeSet<String> = ready_rows
-            .iter()
-            .map(|row| self.logical_queue_name(&row.queue).to_string())
-            .collect();
+        let queues_to_notify: Vec<String> =
+            ready_rows.iter().map(|row| row.queue.clone()).collect();
 
         let mut tx = pool.begin().await.map_err(map_sqlx_error)?;
         let mut total = 0usize;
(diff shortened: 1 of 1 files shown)


hardbyte:
## Wait-event evidence: WAL is the ceiling

Re-ran the same shape with `--wait-events` enabled (single rep, 120 s load). `pg_stat_activity` sampled 1 Hz, non-idle backends aggregated:

| Wait event | Samples | % of active backend time |
|---|---:|---:|
| `LWLock:WALWriteLock` | 239 | **46.4%** |
| `CPU:CPU` | 116 | 22.5% |
| `IO:WalSync` | 107 | **20.8%** |
| `Client:ClientRead` | 27 | 5.2% |
| `Lock:transactionid` | 15 | 2.9% |
| `Lock:relation` | 5 | 1.0% |
| Misc IO/lock | ≤ 3 each | ≤ 0.6% |
| **Total active** | 515 | 100% |

**67% of active backend time is WAL** (`WALWriteLock` + `WalSync`). Application row locks combined are < 4%. CPU is 22.5%. Reducing query-shape round-trips below the WAL ceiling moves zero throughput on a single-host loopback Postgres — fsync, not query work, is the limit.

This matches the README's `awa-longtx-pg-ash` result and explains why this bundle doesn't move `completion_rate` on this bench. The bundle's wins are real but live in:

- **multi-queue NOTIFY fanout** (#233) — bench enqueues to one queue, so 1→1 NOTIFY is identical before/after
- **completion-batch round-trip reduction** (#234) — fixed-cost saving below the WAL noise floor on this hardware
- **enforcing the metric-name invariant via the type system** (#232) — refactor, not perf

## Footprint

Total awa schema size at peak load (30k+ backlogged jobs): **~206 MB**. Hot-table dead-tuple count: zero on every workload-bearing table (ready/lease/claim/done/attempt/dlq/deferred). Auxiliary counters bounded ≤ 174 dead tuples regardless of throughput.

## Disposition

Bundle is throughput-neutral on this bench, dead-tuple-shape preserved, footprint preserved. The three constituent PRs are clean independently — bundle is just review convenience.

hardbyte:
## 128-worker confirmation

Same shape pushed to 128 workers, fixed-rate producer at 8 k/s offered, 90 s load, 2 reps per branch:

| Run | enqueue (med) | completion (med) | completion (peak) |
|---|---:|---:|---:|
| main r1 | 7,999 | 7,998 | 9,807 |
| main r2 | 7,960 | 7,655 | 9,898 |
| bundle r1 | 7,959 | 7,694 | 9,883 |
| bundle r2 | 7,681 | 7,757 | 8,541 |

Mean completion: main 7,827; bundle 7,726. Both branches occasionally clip the 8 k/s offered ceiling, so peaks (~9.9 k/s, identical across branches within 0.2 %) reflect the consumer ceiling. Same parity story as the 64w runs.

Bundle is throughput-neutral at 64 w and 128 w under loopback Postgres — clean refactor + round-trip-saving wins that don't surface against a WAL-bound consumer ceiling on this hardware.

hardbyte:
Pushed a fix on top of the bundle that also lands on the constituent #232: dropped `publish = false` on `awa-metrics` because `awa-worker` (published) depends on the type directly. `cargo publish` strips path deps but keeps the version constraint, so without this `cargo publish -p awa-worker --dry-run` failed with "no matching package named `awa-metrics` found". Verified clean with `--dry-run` on both crates.

PR #240, opened 2026-05-07
Fix cron evaluation starvation

### LINKED ISSUE #239: Cron evaluation can be starved by inline maintenance loop
## Summary

AWA's maintenance leader runs cron evaluation in the same inline `tokio::select!` loop as unrelated maintenance work. If any selected maintenance branch is slow or blocked, cron evaluation is delayed. When evaluation resumes, AWA enqueues only the latest due fire time, so missed fires are collapsed rather than caught up.

This means short-period cron jobs can run much less frequently than configured under load.

## Evidence from code

In `awa-worker/src/maintenance.rs`, the leader loop creates one timer per maintenance task and awaits each selected branch inline:

- `cron_eval_timer.tick()` awaits `self.evaluate_cron_schedules().await`
- `queue_stats_timer.tick()` awaits `self.publish_queue_health_metrics().await`
- `dirty_key_timer.tick()` awaits `self.recompute_dirty_admin_metadata().await`
- `metadata_reconciliation_timer.tick()` awaits `self.refresh_admin_metadata().await`
- 30s/60s rescue, cleanup, sync, leader-check, and priority-aging tasks are also awaited in the same loop

`evaluate_cron_schedules()` computes a single latest due fire per cron row and calls `atomic_enqueue()` once. If the loop is delayed, intermediate due fires are skipped/collapsed.

For example, with a `*/5 * * * * *` schedule, if the maintenance loop is occupied from `12:00:00` until `12:00:35`, the next cron evaluation can enqueue only the latest due fire:

```text
12:00:00 enqueue fire_time=12:00:00
# maintenance loop delayed here
12:00:35 enqueue fire_time=12:00:35
```

There is no catch-up enqueue for `12:00:05`, `:10`, `:15`, `:20`, `:25`, or `:30`.

## Impact

Any short-period AWA cron job can miss expected scheduling windows if an unrelated maintenance task blocks the leader loop. This is especially visible for cron jobs used as polling/reconciliation tasks, where callers may assume a configured 5s/10s cadence.

The current behavior may be acceptable if explicitly documented as "latest fire only", but the starvation from unrelated maintenance work makes the effective cadence unpredictable.

## Related risk: full admin metadata refresh

`refresh_admin_metadata()` performs broad reconciliation work and is called by the same maintenance loop. It also uses invoker privileges and performs:

```sql
TRUNCATE awa.admin_dirty_queues, awa.admin_dirty_kinds;
```

If a runtime role lacks the required privileges, this maintenance branch repeatedly logs an error. If it is slow or blocked for any reason, it can also delay cron evaluation because it is awaited inline.

## Proposed fixes

Options, not mutually exclusive:

1. Run cron evaluation in its own task/loop so unrelated maintenance work cannot starve cron scheduling.
2. Spawn longer maintenance work instead of awaiting it inline in the select loop, with overlap guards where needed.
3. Add duration logging / delayed-tick warnings per maintenance branch so this class of issue is diagnosable from logs.
4. Consider whether cron should catch up missed fires or explicitly document/metric the current "latest fire only" behavior.
5. Review `refresh_admin_metadata()` permissions/security so runtime maintenance does not emit recurring privilege errors.

hardbyte:
## Summary

- run cron evaluation in its own leader-scoped task so unrelated maintenance branches cannot starve schedule evaluation
- preserve existing coalesced/latest-only behavior by default
- add an explicit persisted missed-fire policy: `coalesce` by default, `catch_up` for idempotent reconciliation jobs that need every missed fire in order
- expose the policy through Rust, Python, CLI/API output, docs, and migration v015
- document that the runtime role needs `TRUNCATE` on AWA/control-plane tables and custom queue-storage schemas where runtime maintenance issues TRUNCATE
- add a DB-backed regression test proving the old grants fail with insufficient_privilege and the documented grants allow `refresh_admin_metadata()`

Closes #239

## Testing

- `cargo test -p awa-worker maintenance::tests::compute_fire_times --lib`
- `cargo test -p awa-model cron::tests --lib`
- `cargo check -p awa-cli -p awa-worker`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test -p awa --test migration_test test_documented_runtime_grants_cover_admin_refresh_truncate -- --test-threads=1` compiled locally, but could not execute because local Postgres refused connections
- `cargo check --manifest-path awa-python/Cargo.toml` blocked locally because the available interpreter is Python 3.9 while `awa-python` uses `abi3-py310`

## Release Notes

* **New Features**
  * Added `missed_fire_policy` configuration for cron jobs, controlling how missed fires are handled when the scheduler is delayed. Supports two modes: "coalesce" (default, enqueues only the latest missed fire) and "catch_up" (enqueues all missed fires in order).
  * CLI list output now displays the missed fire policy for each schedule.

* **Documentation**
  * Updated guides and examples with missed-fire policy usage and behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,8 @@ client.periodic(
     "daily_report", "0 9 * * *",
     GenerateReport, GenerateReport(format="pdf"),
     timezone="Pacific/Auckland",
+    # missed_fire_policy="coalesce" by default; use "catch_up" for
+    # idempotent reconciliation jobs that need every missed fire.
 )
 ```
 
--- a/awa-cli/src/main.rs
+++ b/awa-cli/src/main.rs
@@ -661,13 +661,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                             println!("No cron job schedules found.");
                         } else {
                             println!(
-                                "{:<25} {:<20} {:<12} {:<25} {:<10}",
-                                "NAME", "CRON", "TIMEZONE", "KIND", "QUEUE"
+                                "{:<25} {:<20} {:<12} {:<12} {:<25} {:<10}",
+                                "NAME", "CRON", "TIMEZONE", "MISSED", "KIND", "QUEUE"
                             );
                             for s in &schedules {
                                 println!(
-                                    "{:<25} {:<20} {:<12} {:<25} {:<10}",
-                                    s.name, s.cron_expr, s.timezone, s.kind, s.queue,
+                                    "{:<25} {:<20} {:<12} {:<12} {:<25} {:<10}",
+                                    s.name,
+                                    s.cron_expr,
+                                    s.timezone,
+                                    s.missed_fire_policy,
+                                    s.kind,
+                                    s.queue,
                                 );
                             }
                             println!("\n{} schedules listed.", schedules.len());
--- a/awa-model/migrations/v001_canonical_schema.sql
+++ b/awa-model/migrations/v001_canonical_schema.sql
@@ -106,6 +106,8 @@ CREATE TABLE awa.cron_jobs (
     max_attempts     SMALLINT    NOT NULL DEFAULT 25,
     tags             TEXT[]      NOT NULL DEFAULT '{}',
     metadata         JSONB       NOT NULL DEFAULT '{}',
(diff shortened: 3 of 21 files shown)


hardbyte:
@coderabbitai review

PR #250, opened 2026-05-10
ADR-024: deferred done_entries — investigated, rejected

hardbyte:
## Status: REJECTED — revert committed in 053fec1

Investigated, prototyped, benchmarked, and backed out. The synchronous receipt-ring complete path stays the only first-class implementation.

## What the bench actually showed

Same hardware as the 2026-05-09 sweep shape — producer-rate=50000, depth-target=4000, W=64, 30 s warmup + 60 s clean — local docker pg18:

| mode | completion (jobs/s) | enqueue | depth | e2e p99 |
|---|---:|---:|---:|---:|
| sync (this branch) | **2,803** | 3,222 | 4,140 | 1,396 ms |
| deferred (final design) | 1,839 | 2,023 | 3,895 | 1,954 ms |

Deferred shipped **34 % lower** completion-rate and **40 % higher** p99. The first-attempt design (closure prune gated on materialiser progress) was much worse — multi-second stalls — and prompted a redesign that inlined the materialise pass in \`prune_oldest_claims\` itself. That redesign removed the stalls but didn't recover throughput parity, let alone deliver lift.

## Why it didn't pan out

- **The \`done_entries\` INSERT is not the bottleneck on this workload.** Sync mode at 2.8 k jobs/s on macOS docker / pg18 leaves plenty of headroom; the JSONB write isn't where the time goes.
- **The deferred design adds three coupled mechanisms** (periodic materialiser timer, inline catch-up in \`prune_oldest_claims\`, rotation guard via the existing pending-ready check). Together they introduce partition-rotation contention that exceeds the saved hot-path WAL.
- **The read-side migration was still pending** (\`v016\` UNION branch in \`awa.jobs_compat()\`). The ADR couldn't flip the default without it.

The simpler architecture wins. The throughput gap to pgque (14 k vs 40 k in the sweep) is structural — pgque trades feature surface and per-job durability for throughput, awa keeps both. The right awa improvements live elsewhere — the bloat-resistance drops under \`idle_in_tx\` and \`active_readers\` (75 % / 71 %) are awa-specific and closeable; that's where the next investigation should look.

## What's kept

- **ADR-024**, status flipped to Rejected, with the bench data recorded — captures the design space and why it didn't pan out.
- **\`correctness/storage/AwaDeferredMaterialize.tla\`**, with a header comment noting rejection so future readers don't assume it lives in code. Useful as a record of the safety + liveness contract any future revisit would need to satisfy.
- **The two priority-aging tests** ([awa#249](https://github.com/hardbyte/awa/pull/249) — separate PR; this branch is a strict superset).

## What's removed

\`QueueStorageConfig.deferred_done_entries\` field, \`complete_runtime_batch\`'s deferred branch, \`materialise_done_entries\` and \`unmaterialised_closures_for_ready_partition\` SQL helpers, \`prune_oldest_claims\`'s inline-materialise, the maintenance loop's materialise timer, the 5 ADR-024 integration tests, the \`release_unique_claims_for_done_rows_tx\` helper extraction (inlined back), and the bulk-added \`deferred_done_entries: false\` lines across the test suite.

## Test plan

- [x] \`cargo build --workspace\` clean
- [x] \`cargo clippy --workspace --tests -- -D warnings\` clean
- [x] \`cargo fmt --all\`
- [x] \`cargo test -p awa --test queue_storage_runtime_test\` — 57/57 pass (was 63/63 with the deferred-mode tests)
- [x] A/B benchmark on local docker pg18 — see numbers above

### DIFF
--- a/awa-model/src/queue_storage.rs
+++ b/awa-model/src/queue_storage.rs
@@ -41,21 +41,6 @@ pub struct QueueStorageConfig {
     /// Set to `false` to force every claim through the legacy
     /// `leases` materialization path.
     pub lease_claim_receipts: bool,
-    /// ADR-024: defer the `done_entries` write off the completion
-    /// hot path. When `true`, completing a job writes only the
-    /// `lease_claim_closures` row and `attempt_state` delete; a
-    /// periodic materialiser task joins closures with `lease_claims`
-    /// and `ready_entries` to backfill `done_entries`. The read-side
-    /// `jobs_view` UNIONs in not-yet-materialised closures so
-    /// callers never observe a stale state. The
-    /// `ready_entries`-partition rotation guard blocks TRUNCATE
-    /// while any closure still references the partition without a
-    /// corresponding `done_entries` row.
-    ///
-    /// Default `false` — pre-flighted under env flag
-    /// `AWA_DEFER_DONE_ENTRIES=1` until validation lands. See ADR-024
-    /// and `correctness/storage/AwaDeferredMaterialize.tla`.
-    pub deferred_done_entries: bool,
 }
 
 impl Default for QueueStorageConfig {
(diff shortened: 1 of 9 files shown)


hardbyte:
Closing without merging. ADR-024 is captured in the PR description and commit history (52dc713..7cffa48) for future search. The TLA+ model is over-engineered preservation for a rejected design — a fresh spec would be needed for any future revisit anyway.

PR #253, opened 2026-05-12
perf(queue_storage): in-process ensure_lane cache + completion-batcher defaults

hardbyte:
## Summary

- **+62% throughput** on a 16-producer same-queue queue-storage enqueue workload (measured: 29,793 → 48,404 jobs/sec across three local A/B runs)
- In-process cache for `(queue, priority)` lane presence on `QueueStorage` — skips three `INSERT ... ON CONFLICT DO NOTHING` round-trips on subsequent enqueue batches for an already-established lane
- Completion-batcher defaults move from `(batch=128, flush=1ms)` to `(batch=256, flush=5ms)`; tunable via `AWA_COMPLETION_BATCH_SIZE` / `AWA_COMPLETION_FLUSH_MS`
- New in-tree `test_queue_storage_enqueue_contention` so queue-storage enqueue contention has an A/B fixture (the existing matrix only exercised canonical `jobs_hot`)

No public API change. No schema change. No semantic change. Default behaviour is observationally identical for callers — the cache is correctness-preserving via an `advance_enqueue_head` retry path that detects a rolled-back ensure_lane and re-runs it.

## Why this matters now

The 2026-05-09 portable-bench sweep flagged `UPDATE queue_enqueue_heads` as the slowest hot-path query (mean 14.8–18.0 ms at 256 producers). A wait-event probe on a 16-producer local repro attributed 93% of producer time to `Lock:transactionid` / `Lock:tuple`, of which 93% was on the `enqueue_head_update` query — i.e. ~87% of total producer time on a single row.

The three `INSERT ... ON CONFLICT DO NOTHING` round-trips that establish the lane row, the head row, and the claim row aren't the dominant cost, but they're cheap to eliminate after the first call per `(queue, priority)`. The completion-batcher tuning is a separate small lever for the completion side.

## Cache correctness

The cache is *optimistic*: it marks a lane "ensured" inside the same transaction that ran the three INSERTs. If that transaction ultimately rolls back, the rows vanish but the cache still claims they exist. The subsequent enqueue would then fail when its `UPDATE queue_enqueue_heads ... RETURNING` finds zero rows.

`advance_enqueue_head` handles this:

1. `UPDATE ... RETURNING` with `fetch_optional`.
2. If `Some(start_seq)` — return it.
3. If `None` — invalidate the cached lane, re-run `ensure_lane`, retry the `UPDATE` exactly once with `fetch_one`.

Reset clears the cache because `reset()` TRUNCATEs `queue_lanes`.

A new regression test `test_queue_storage_ensure_lane_cache_recovers_after_rollback` exercises this path by deleting the head row out from under the cache, then confirming a subsequent enqueue self-heals.

## Validation

- `cargo fmt --all` — no diff
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test -p awa --test queue_storage_runtime_test` — **58 passed**
- Local A/B (3 runs, 16 producers x 15k jobs, same queue): 29,793 → 48,404 jobs/sec (+62%)

## Test plan

- [x] `cargo test -p awa --test queue_storage_runtime_test`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] Local A/B bench with `test_queue_storage_enqueue_contention`
- [ ] CI run-through on this branch

## Follow-up (not in this PR)

The companion Phase D PR (#TBD) builds on this branch to shard `queue_enqueue_heads` by an operator-tunable `enqueue_shards`. That delivers a further 4.84× on the same workload at S=8 (200k jobs/sec), but it carries a schema migration + a FIFO-per-shard semantic change, so it ships separately for review.

## Summary by CodeRabbit

* **Performance**
  * Added in-process caching to reduce database round-trips during queue enqueues.
  * Increased default completion batch size and flush interval; defaults are now larger and remain configurable.

* **Reliability**
  * Improved cache invalidation and recovery so enqueues remain correct after rollback scenarios.

* **Tests**
  * Added benchmark and integration tests to validate enqueue contention and cache-recovery behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/hardbyte/awa/pull/253)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,19 @@ transitions live in [`docs/upgrade-0.5-to-0.6.md`](docs/upgrade-0.5-to-0.6.md).
   source under pinned-xmin workloads (long-running reader
   transactions) without changing public API behaviour. See ADR-019
   § `lane_state` and segment cursor tables.
+- Cache `(queue, priority)` lane presence in-process on the
+  queue-storage path so subsequent enqueue batches skip the three
+  `INSERT ... ON CONFLICT DO NOTHING` round-trips on
+  `queue_lanes` / `queue_enqueue_heads` / `queue_claim_heads`. The
+  cache invalidates and re-runs `ensure_lane` if a subsequent
+  `UPDATE queue_enqueue_heads` finds no row (the observable signal
+  of an earlier ensure_lane that ran inside a rolled-back
+  transaction), so correctness is preserved.
+- Raise the completion-batcher defaults to `(batch=256, flush=5ms)`
+  from `(batch=128, flush=1ms)`. Lets batches amortise the per-batch
+  completion SQL over more rows at the cost of a small latency
+  bump. Tunable via `AWA_COMPLETION_BATCH_SIZE` and
+  `AWA_COMPLETION_FLUSH_MS`.
 
 ## [0.6.0-alpha.9] — 2026-05-08
 
--- a/awa-model/src/queue_storage.rs
+++ b/awa-model/src/queue_storage.rs
@@ -3781,12 +3781,27 @@ impl QueueStorage {
         // rows in some transaction. The cache is optimistic — if that
         // earlier transaction rolled back, the head row is missing
         // even though the cache says it exists. `advance_enqueue_head`
-        // detects that case via the empty UPDATE result and invalidates
-        // the cache entry before retrying.
+        // detects that case via the empty UPDATE result, invalidates
+        // the cache entry, and re-runs `ensure_lane_inserts` directly
+        // (bypassing the fast path).
         if self.lane_is_cached(queue, priority) {
             return Ok(());
         }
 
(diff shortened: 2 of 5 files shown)


hardbyte:
## Review pass — 2026-05-13

**P2: cache recovery race on stale-cache invalidation** — fixed in `68abe63`.

The recovery path in `advance_enqueue_head` previously called `ensure_lane` after invalidating, which would re-take the cache fast path if a concurrent transaction had re-marked the same lane between the invalidate and the check (e.g. its `mark_lane_ensured` ran but its INSERTs hadn't committed yet). The retry UPDATE would then still find no row.

Split `ensure_lane` into two helpers:

- `ensure_lane` keeps the fast-path check and is the normal entry point.
- `ensure_lane_inserts` runs the three INSERTs unconditionally. The recovery path calls this directly so it cannot be tricked into skipping the repair.

The existing rollback-recovery test continues to pass; its docstring now explains the concurrent shape and points at `advance_enqueue_head`'s comment for the invariant.

PR #266, opened 2026-05-19
Fix queue-storage prepare_schema startup race

### LINKED ISSUE #264: prepare_schema concurrent-startup race on fresh DB (especially PG18)
## Symptom

On a fresh queue-storage schema (no `awa.job_id_seq` yet), multiple worker pods starting concurrently all enter `prepare_schema` at the same time. On PG18 in particular the concurrent `CREATE SEQUENCE IF NOT EXISTS awa.job_id_seq` blocks indefinitely — no errors, no active queries visible in `pg_stat_activity`, just \"Starting Awa worker runtime\" log lines forever. Any producer that starts in parallel hits:

```
database error: error returned from database: relation "awa.job_id_seq" does not exist
```

…and stays stuck in its backoff loop until the consumer race resolves.

We hit this twice in the same staging run:

1. First-ever fresh-DB setup on Cloud SQL `discovery-staging` (PG18) — 4 consumer pods all stuck for >5 min until a manual `psql -c "CREATE SEQUENCE IF NOT EXISTS awa.job_id_seq"` from outside unblocked them, after which a consumer rollout completed `prepare_schema` cleanly in ~600 ms.
2. After a mid-bench `DROP SCHEMA awa CASCADE` + remigrate, the existing consumer pods didn't redo `prepare_schema` (they cached state across the schema drop) — so producers immediately hit \"relation awa.job_id_seq does not exist\" and the next consumer rollout was needed to recreate it.

## Why it's worth fixing

- It's a silent failure: no error from PG, no error from the runtime, just an unbounded wait on PG18.
- It hits every fresh-DB deploy of awa, not just ours — anyone bringing up the runtime against a clean Postgres for the first time will see it if they spin up more than one pod simultaneously, which is the default for any Deployment with `replicas > 1`.
- The mid-flight version (case 2) is a footgun for anyone doing schema reset (e.g. recovery-from-corruption runbooks, dev-env reset scripts).

## Suggested fix options

1. **Move sequence + ring-state + claim-head DDL into a numbered migration.** The migration runner is already serialised by `awa.schema_version`, so no race. This is the cleanest answer and probably the right long-term shape — `prepare_schema` then only handles things that genuinely depend on per-runtime configuration (slot count, etc.).

2. **Take an advisory lock at the top of `prepare_schema`.** Smaller change; keeps the dual-purpose schema setup. `pg_advisory_xact_lock(hashtext('awa.prepare_schema'))` would serialise without blocking other workloads.

If you go with option 1, the touch points are around `awa-model/src/queue_storage.rs:3284,3317,3352` (and the surrounding `prepare_schema` calls a few lines above each) — most of the `CREATE SEQUENCE` / `CREATE TABLE IF NOT EXISTS` calls for the queue-storage substrate. PR #263 (just merged) updated those three index sites; the prepare_schema function still contains the bulk of the DDL.

## Reproducer (suggested)

A test in `awa/tests/` that:

1. `DROP SCHEMA awa CASCADE` on a fresh PG18 testcontainer.
2. Run migrations to head.
3. Spin up N=4 concurrent worker startups (each calling `QueueStorage::new` + `worker.run()`).
4. Have one producer task also start in parallel, calling `enqueue_many_copy` immediately.
5. Assert all 5 succeed inside a 30-second deadline.

On `main` today this would either hang or surface the \"relation awa.job_id_seq does not exist\" producer error.

## Workaround in the meantime

Either:
- Bring the runtime up with a single consumer pod first, let it finish `prepare_schema`, then scale to the target replica count.
- Or run a one-shot `CREATE SEQUENCE IF NOT EXISTS awa.job_id_seq` from psql as part of the deploy pipeline before consumer pods come up.

Neither is acceptable for a self-deploying runtime.

hardbyte:
## Summary
- run queue-storage prepare_schema under one transaction-scoped advisory lock instead of holding one pooled lock connection while the install body reacquires the pool
- tighten queue-storage schema readiness so a missing job_id_seq or claim runtime function is not treated as ready
- add regressions for one-connection prepare_schema and concurrent prepare_schema startup

Fixes #264.

## Tests
- cargo check -p awa-model
- cargo test --test queue_storage_runtime_test test_queue_storage_prepare_schema -- --nocapture
- cargo test --test migration_test test_queue_storage_schema_ready_requires_sequence_and_claim_function -- --nocapture
- cargo fmt --check
- git diff --check

### DIFF
--- a/awa-model/src/queue_storage.rs
+++ b/awa-model/src/queue_storage.rs
@@ -2008,17 +2008,23 @@ impl QueueStorage {
     pub async fn prepare_schema(&self, pool: &PgPool) -> Result<(), AwaError> {
         let schema = self.schema();
         let install_lock_name = format!("awa.queue_storage.install:{schema}");
-        let mut install_lock_conn = pool.acquire().await.map_err(map_sqlx_error)?;
+        let mut install_tx = pool.begin().await.map_err(map_sqlx_error)?;
 
-        sqlx::query("SELECT pg_advisory_lock(hashtextextended($1, 0))")
+        sqlx::query("SELECT pg_advisory_xact_lock(hashtextextended($1, 0))")
             .bind(&install_lock_name)
-            .execute(install_lock_conn.as_mut())
+            .execute(install_tx.as_mut())
             .await
             .map_err(map_sqlx_error)?;
 
+        macro_rules! install_db {
+            () => {
+                install_tx.as_mut()
+            };
+        }
+
         let install_result = async {
             sqlx::query(&format!("CREATE SCHEMA IF NOT EXISTS {schema}"))
-                .execute(pool)
+                .execute(install_db!())
                 .await
                 .map_err(map_sqlx_error)?;
 
@@ -2041,14 +2047,14 @@ impl QueueStorage {
                 "#,
             )
             .bind(schema)
-            .fetch_one(pool)
+            .fetch_one(install_db!())
             .await
             .map_err(map_sqlx_error)?;
             if open_receipt_claims_exists {
                 let row_count: i64 = sqlx::query_scalar(&format!(
                     "SELECT count(*)::bigint FROM {schema}.open_receipt_claims"
                 ))
-                .fetch_one(pool)
+                .fetch_one(install_db!())
                 .await
                 .map_err(map_sqlx_error)?;
                 if row_count > 0 {
@@ -2062,7 +2068,7 @@ impl QueueStorage {
                 sqlx::query(&format!(
                     "DROP TABLE IF EXISTS {schema}.open_receipt_claims CASCADE"
                 ))
-                .execute(pool)
+                .execute(install_db!())
                 .await
                 .map_err(map_sqlx_error)?;
             }
@@ -2198,7 +2204,7 @@ impl QueueStorage {
                 CREATE SEQUENCE IF NOT EXISTS {schema}.job_id_seq
                 "#
             ))
-            .execute(pool)
+            .execute(install_db!())
             .await
             .map_err(map_sqlx_error)?;
 
@@ -2212,7 +2218,7 @@ impl QueueStorage {
             )
             "#
             ))
-            .execute(pool)
+            .execute(install_db!())
             .await
             .map_err(map_sqlx_error)?;
 
@@ -2232,7 +2238,7 @@ impl QueueStorage {
             )
             "#
             ))
-            .execute(pool)
+            .execute(install_db!())
             .await
             .map_err(map_sqlx_error)?;
 
@@ -2244,7 +2250,7 @@ impl QueueStorage {
             "#
             ))
             .bind(self.queue_slot_count() as i32)
-            .execute(pool)
+            .execute(install_db!())
             .await
             .map_err(map_sqlx_error)?;
 
@@ -2256,7 +2262,7 @@ impl QueueStorage {
             )
             "#
             ))
-            .execute(pool)
+            .execute(install_db!())
             .await
             .map_err(map_sqlx_error)?;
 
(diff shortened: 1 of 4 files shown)


hardbyte:
## LGTM — right diagnosis, right fix, well-targeted tests

(Posting as a comment rather than a review since `gh` won't let the author self-approve. Treat as a +1 from another set of eyes.)

The bug is exactly what the test suggests: the old code held `pg_advisory_lock(…)` on a single dedicated connection but ran every DDL through `.execute(pool)`, which acquires *different* connections from the pool each time. With a one-connection pool that deadlocks immediately (the lock-holder connection is the only one available, but it's blocked waiting for itself); with a small pool under concurrent startup you get pool exhaustion (every pod holds one connection for its lock and competes for the rest for DDL). That's the staging hang in #264 and the `one-connection-pool` regression test reproduces it cleanly.

Switching to `pg_advisory_xact_lock(…)` on a single transaction-bound connection (`install_tx`) is the right architectural move. All DDL now runs on the same connection inside the same transaction; the lock auto-releases on COMMIT/ROLLBACK and the install is now atomic, which incidentally upgrades the failure-mode semantics — previously a mid-install failure left a partially-installed schema, now it all rolls back.

Tightening `queue_storage_schema_ready` to require `job_id_seq`, the other queue-storage tables, and the `claim_ready_runtime` function is a nice complementary improvement. Even with the lock fix in place, an aborted install (e.g. transaction rollback after a partial failure on the old code path) used to leave `queue_ring_state` + `ready_entries` + `leases` present but `job_id_seq` missing — exactly the state I saw in staging where producers were hitting `relation "awa.job_id_seq" does not exist` while consumers were stuck. Workers calling `schema_ready` can now trust the result.

## Observations

A few small things — none are blockers and I'd merge as-is.

1. **`install_db!()` macro vs a helper function.** The macro just expands to `install_tx.as_mut()`. It's the right call (each invocation re-borrows independently within the same scope, which the borrow checker prefers over a helper that returns `&mut Tx<…>`). Worth a one-line comment on the macro saying that's why it isn't a function; future readers will reach for the function rewrite reflexively.

2. **Lock duration covers all CREATE INDEX calls.** Since none of them are `CONCURRENTLY` (which Postgres bans inside a transaction anyway), they take `ShareUpdateExclusiveLock` on each table for the duration of the index build. On fresh installs that's a no-op cost. On a redeploy of an existing system with rows already in `ready_entries`, the build can take meaningful time and other sessions writing to that table will queue up. Already true on `main`; not a regression. Worth a sentence in the prepare_schema doc-comment so operators know.

3. **Concurrent-startups serialize test.** Good test — exactly what I'd want. The wall-clock is now `N × time_for_one_prepare` because they serialize on the advisory lock; the 30-second timeout is comfortable for fresh setup but if `prepare_schema` ever grows (more partitions, more indexes) this could get tight. Fine as is.

4. **Advisory lock key namespace.** `hashtextextended('awa.queue_storage.install:<schema>', 0)` is well-namespaced. The only collision risk is with another awa subsystem that also calls `pg_advisory_xact_lock` with a hashtext-style key, which doesn't exist today. If anyone adds another advisory-lock-protected install path, they should pick a different prefix.

5. **Readiness check test names a `prepare_queue_storage_schema` helper.** I assume that exists in the test crate already (both new tests rely on it). Worth a quick scan to confirm it's not a missing helper.

6. **Optional: include `claim_ready_runtime` source-of-truth pointer in the doc-comment.** The readiness check's signature `(text, bigint, double precision, double precision)` mirrors the function definition in `prepare_schema`. If the signature ever changes you'd need to update both places; a comment on the readiness check pointing at the definition site (or vice-versa) would catch that.

Compile-checked locally with `SQLX_OFFLINE=true cargo check --workspace` and `cargo fmt --all -- --check` — both clean.

Nice fix. Once CI is green I'd merge.

Issue #278, opened 2026-05-31
Extract shared callback contract and signature helpers


hardbyte:
## Context

ADR-027 proposes treating callback ingress as a deployable surface separate from the admin UI/API. The first implementation step is to stop owning the callback HTTP contract inside `awa-ui`.

Today the worker-side signer lives in `awa-worker/src/http_worker.rs`, while the receiver-side verifier and payload parsing live in `awa-ui/src/handlers/callbacks.rs`. That makes custom receivers easy to drift from the built-in one.

Related design draft: #277.

## Proposed work

Extract the shared callback contract and auth helpers into a module or crate that can be used by the worker, the current UI callback routes, and future callback-only routers.

Include:

- `X-Awa-Signature` header constant
- BLAKE3 keyed-hash signing and verification helpers
- 32-byte callback secret parsing
- typed request/response shapes for `complete`, `fail`, and `heartbeat`
- heartbeat timeout validation
- common error mapping suitable for HTTP handlers

The existing option name `hmac_secret` can remain for compatibility, but docs and APIs should be explicit that the algorithm is BLAKE3 keyed hashing, not RFC HMAC.

## Acceptance criteria

- `HttpWorker` and the existing `awa-ui` callback handlers use the same signing/verifying implementation.
- Tests cover valid, missing, malformed, and incorrect signatures.
- Tests cover complete/fail/heartbeat payload parsing and invalid heartbeat timeouts.
- The no-secret behavior is explicit in configuration/tests rather than falling out accidentally from `Option::None`.
- Public docs continue to describe the wire contract accurately.

Issue #281, opened 2026-05-31
Document and support user-owned callback API layers


hardbyte:
## Context

Some users will not want Awa to own the HTTP server that receives callbacks. They may already have FastAPI, Starlette, Django, Flask, axum, or another API layer and want to expose Awa callback routes inside that application.

Today the docs say users can mount equivalent handlers around admin functions, but that leaves them to copy the signing algorithm, payload contract, timeout validation, and error behavior.

Related design draft: #277. Shared callback contract/auth is tracked in #278.

## Proposed work

Provide an officially supported integration path for user-owned callback API layers.

Rust direction:

- expose lower-level callback receiver service functions that are not tied to axum routing;
- document how to call them from an existing axum application;
- make the axum callback-only router from #279 a thin wrapper around those functions.

Python direction:

- expose or document a small helper for verifying `X-Awa-Signature`;
- provide FastAPI-style examples for complete/fail/heartbeat routes;
- show how to resolve callbacks via the Python admin/client API;
- keep the examples aligned with the shared contract from #278.

## Acceptance criteria

- Docs include a minimal custom axum example.
- Docs include a minimal FastAPI example.
- Examples do not reimplement the signature algorithm inline when a helper exists.
- The examples cover complete, fail, and heartbeat.
- The security docs clearly state that callback routes are mutating endpoints and must be signed or otherwise protected.
- If Python helpers are added, they have parity tests against the Rust signing implementation.

hardbyte:
Will have another pass to ensure all changes have landed in the right place (eg bit in awa-ui crate) and are well documented and tested

hardbyte:
Closing as implementation-complete. #293 shipped the callback-only router and user-owned helpers, with axum and FastAPI documentation and Python helper parity tests. The shared callback contract and configurable URL prefix landed via #291.

PR #292, opened 2026-06-01
feat(callbacks): callback-only receiver router and CLI (#279)

### LINKED ISSUE #279: Add a callback-only receiver router and CLI
## Context

`awa serve` currently exposes the admin UI, admin REST API, static fallback, permissive UI CORS, and HTTP-worker callback endpoints from one router. That is awkward for deployments where callbacks must be externally reachable but the admin surface must remain private.

ADR-027 proposes making callback ingress a first-class deployable surface. Related design draft: #277. Shared contract/auth extraction is tracked in #278.

This is related to, but narrower than, the broader `awa-api` / `awa-ui` split in #143.

## Proposed work

Add a callback-only receiver surface that can be run without the UI or admin API.

Possible public shapes:

```rust
let router = awa_callbacks::router(pool, CallbackReceiverConfig {
    secret: Some(callback_secret),
    path_prefix: "/api/callbacks".into(),
    ..Default::default()
});
```

and:

```text
awa callbacks serve
```

The CLI should accept database/pool options plus host, port, callback signing secret, path prefix, and an explicit unsigned/private-network escape hatch if we keep that mode.

## Acceptance criteria

- Callback-only router mounts only complete/fail/heartbeat routes.
- It does not serve static UI assets.
- It does not expose admin REST routes.
- It does not apply permissive UI CORS by default.
- It requires writable database access, since all callback routes mutate job state.
- It requires a callback signing secret by default, unless explicitly configured otherwise.
- The existing `awa serve` callback paths remain compatible.
- Router/CLI tests cover success and auth failure for all callback routes.
- `docs/http-callbacks.md` and `docs/security.md` describe when to use callback-only serving vs `awa serve`.

hardbyte:
## Summary

Adds a callback-only deployable role — a router that exposes ONLY the three callback ingress endpoints, plus a CLI subcommand to run it standalone:

```bash
awa callbacks serve \
  --host 0.0.0.0 --port 4000 \
  --callback-hmac-secret \"\$AWA_CALLBACK_HMAC_SECRET\"
```

The library entry point is `awa_ui::callback_router(pool, CallbackReceiverConfig)`.

## Stacked on #291

This builds on the shared callback contract in #291 (#278/#280) — the router uses `awa_model::callback_contract::DEFAULT_CALLBACK_PATH_PREFIX` and the URL/auth helpers introduced there. Review #291 first; this branch's base is `brian/issue-278-callback-contract` and will retarget `main` when that lands.

## What the receiver does NOT do

By design, distinguishing it from `awa serve`:
- No static UI assets
- No admin REST routes (jobs, queues, DLQ, runtime, stats)
- No permissive CORS layer
- Refuses to build against a read-only pool (every route mutates job state — fail fast at build time instead of surfacing as runtime 503s)

## CallbackAuth is an explicit enum

```rust
pub enum CallbackAuth {
    Signed([u8; 32]),
    Unsigned,
}
```

So the security choice cannot be made by forgetting a setter. `Unsigned` is intentionally awkward — choose it only for trusted-network deployments (mTLS at the load balancer, IP allow-list, private VPC).

The CLI surface enforces this: `--callback-hmac-secret` and `--allow-unsigned` are mutually exclusive, and neither defaults to a permissive mode. A clear error fires when both are omitted:

```text
Error: \"a callback signing secret is required by default; pass --callback-hmac-secret <hex32> or, for a trusted-network deployment, --allow-unsigned\"
```

## Closes

- Closes #279

Refs #277 (design draft), ADR-027.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo build --workspace --all-targets`
- [x] `cargo test -p awa-ui --test callback_router_test` against Postgres — 5 tests:
  - `callback_router_complete_requires_valid_signature` (valid signature + missing-signature 401)
  - `callback_router_fail_and_heartbeat_routes` (signed POSTs to the other two routes)
  - `callback_router_does_not_expose_admin_or_static` (404s on `/api/stats`, `/api/jobs`, `/`, etc.)
  - `callback_router_unsigned_mode_accepts_unauthenticated_requests`
  - `callback_router_custom_path_prefix` (custom mount; default 404s)
- [x] `cargo test -p awa-ui --test callback_api_test` — original 2 admin-router callback tests still green
- [x] `awa callbacks serve --help` — flags documented
- [x] CLI smoke: secret-and-unsigned-conflict and neither-supplied both fail with clear messages
- [x] `docs/http-callbacks.md` — new "Deploying callback ingress separately" section

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -510,6 +510,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sqlx",
+ "thiserror 2.0.18",
  "tokio",
  "tower",
  "tower-http",
--- a/awa-cli/src/main.rs
+++ b/awa-cli/src/main.rs
@@ -107,6 +107,60 @@ enum Commands {
         #[arg(long, env = "AWA_READ_ONLY")]
         read_only: bool,
     },
+    /// Callback receiver subcommands
+    Callbacks {
+        #[command(subcommand)]
+        command: CallbackCommands,
+    },
+}
+
+#[derive(Subcommand)]
+enum CallbackCommands {
+    /// Start a callback-only receiver (no admin UI, no admin API).
+    ///
+    /// Use this when callbacks must be externally reachable but the admin
+    /// surface must remain private. See ADR-027 and docs/http-callbacks.md.
+    Serve {
+        /// Host to bind to
+        #[arg(long, default_value = "127.0.0.1")]
+        host: String,
+        /// Port to listen on
+        #[arg(long, default_value = "4000")]
+        port: u16,
+        /// Maximum number of database connections
+        #[arg(long, default_value = "10", env = "AWA_POOL_MAX")]
+        pool_max: u32,
+        /// Minimum idle connections kept open
+        #[arg(long, default_value = "2", env = "AWA_POOL_MIN")]
+        pool_min: u32,
+        /// Seconds before an idle connection is closed
+        #[arg(long, default_value = "300", env = "AWA_POOL_IDLE_TIMEOUT")]
+        pool_idle_timeout: u64,
+        /// Maximum lifetime of a connection in seconds
+        #[arg(long, default_value = "1800", env = "AWA_POOL_MAX_LIFETIME")]
+        pool_max_lifetime: u64,
+        /// Seconds to wait when acquiring a connection
+        #[arg(long, default_value = "10", env = "AWA_POOL_ACQUIRE_TIMEOUT")]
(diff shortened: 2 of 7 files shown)


hardbyte:
Closed because the base branch `brian/issue-278-callback-contract` was deleted on the #291 merge. The work in this PR (the #279 callback-only router + CLI) is now being landed via #293, which contains both the #279 and #281 commits atop main.

Issue #290, opened 2026-06-01
perf(queue_storage): live terminal counter for queue_counts_exact


hardbyte:
## Context

PR #289 added `QueueStorage::queue_counts_fast` — an index-only depth probe for high-cadence pollers (admin dashboards, depth-target producers, soak observability). It is *not* a replacement for `queue_counts_exact`: it under-counts `completed` while the live `done_entries` segment hasn't rolled up, and the exact path is still what admin tooling reaches for when the live terminal segment matters.

`queue_counts_exact` itself stays slow at deep backlog. Two CTEs dominate:

- `live_terminal`: `count(*) FROM done_entries WHERE queue = $1`. The per-slot `done_entries` child tables have a composite index leading with `(queue, priority, enqueue_shard, lane_seq)`, so a queue-only filter can in principle index-only-scan — but that still has to count every matching terminal row, and append-heavy recent rows aren't reliably index-only until vacuum marks pages all-visible.
- `live_running`'s `lease_claims × lease_claim_closures` anti-join — separate problem, not in scope here.

A naive `done_entries(queue)` index would mostly duplicate the leading columns of the existing composite index. It might cut buffer reads, but it's still O(live terminal rows) per call.

The better shape is **an exact rollup that makes `queue_counts_exact` O(queue slots × shards), not O(done rows)**.

## Proposal

Add a live terminal counter alongside `queue_terminal_rollups.pruned_completed_count`, keyed naturally on the same dimensions the hot path already partitions on:

| Counter | Key shape |
|---|---|
| `queue_terminal_live_counts.completed` | `(ready_slot, queue, priority, enqueue_shard)` |

(Or `(ready_slot, queue, priority)` if `enqueue_shard` granularity isn't worth the extra rows.)

**Write paths to update**:

- `insert_done_rows_tx` — group by `(ready_slot, queue, priority, enqueue_shard)` and `INSERT … ON CONFLICT … DO UPDATE SET completed = completed + EXCLUDED.completed` once per group per batch.
- Terminal row removal: `retry_job_tx` (terminal → retryable), `move_failed_to_dlq` (terminal → DLQ), any other terminal-deleting path. Decrement by the moved count.

**Prune path**: when a `ready_slot` is pruned, fold its live counters into `queue_terminal_rollups.pruned_completed_count` for that `(queue, priority)` (sum or max, matching the existing aggregator), then `DELETE FROM queue_terminal_live_counts WHERE ready_slot = $slot`.

**Exact query becomes**:

[...]

Issue #242, opened 2026-05-07
Isolate user-visible maintenance lanes from background hygiene


hardbyte:
## Summary

AWA's maintenance leader still runs most non-cron maintenance branches in one inline `tokio::select!` loop. PR #240 isolates cron evaluation because it is user-visible scheduling, but other user-visible maintenance work can still be delayed by unrelated hygiene work.

This issue tracks a follow-up refactor to isolate the remaining user-visible lanes without turning #240 into a broad maintenance scheduler rewrite.

## Why this matters

Some maintenance branches affect user-space behavior directly:

- `promote_scheduled()` makes delayed/retryable jobs available.
- `rescue_stale_heartbeats()` detects crashed workers.
- `rescue_expired_deadlines()` enforces hard deadlines.
- `rescue_expired_callbacks()` resumes timed-out callback waits.
- queue-storage ring rotation/pruning affects storage health and may use `TRUNCATE`.

These should not be delayed by slower admin/cache/metrics/cleanup work such as queue-health metric publication, dirty-key recompute, full admin metadata refresh, descriptor cleanup, or completed-job cleanup.

## Proposed approach

1. Split user-visible lanes into separate leader-owned cancellable tasks:
   - scheduled promotion
   - heartbeat/deadline/callback rescue
   - queue-storage queue/lease/claim rotation
2. Add explicit overlap guards where a branch must not run concurrently with itself.
3. Keep admin hygiene in lower-priority loops:
   - queue stats publication
   - dirty metadata recompute
   - full admin metadata refresh
   - cleanup of completed/DLQ/runtime-snapshot/descriptor rows
4. Add per-branch duration logging and delayed-tick warnings so future starvation is visible in logs/metrics.
5. Preserve leader-election semantics: tasks start only while this process owns the maintenance advisory lock and stop promptly when leadership is lost or the client shuts down.

## Acceptance criteria

- User-visible maintenance lanes cannot be blocked by admin hygiene branches.
- Shutdown and leadership-loss cancellation stop all spawned maintenance tasks.
- Overlap-sensitive branches are guarded against concurrent self-overlap.
- Tests cover at least one slow hygiene branch not delaying a user-visible lane, plus cancellation/leadership-loss cleanup for spawned tasks.
- Docs/architecture note the split between user-visible maintenance and background hygiene.

## Related

- Follow-up from #239 and #240.

hardbyte:
## Release decision for 0.6.0

[...]

Issue #169, opened 2026-04-18
Validate MVCC / dead-tuple behaviour under sustained load


hardbyte:
## Context

Every Postgres-backed job queue that uses `SELECT ... FOR UPDATE SKIP LOCKED` with per-row UPDATE/DELETE for state transitions creates dead tuples on the hot table. Under sustained load with long-running transactions pinning the MVCC horizon, these dead tuples accumulate faster than autovacuum can reclaim them. Indexes bloat, SKIP LOCKED scans walk more dead entries, throughput degrades, back-pressure builds, and the queue enters a documented death spiral.

Important framing: this is not inherent to Postgres-backed queues. It is inherent to the per-row-state-machine storage model. See the [design-space split](#design-space-split) below. Awa's current (0.5.x) implementation uses that model because it is the simpler path to the feature set (heartbeats, DLQ, per-job cancel, priorities). Nothing in the public API forces us to stay there. A future version could offer rotation-and-TRUNCATE as an opt-in storage mode for append-heavy queues where the richer per-row semantics are not needed. This issue tracks both the measurement of where we are today and the decision of whether to add that second mode.

## Relationship to #165 and #166

This issue assumes the measurement infrastructure in #165 and #166 is in place. #165 builds the portable long-horizon scenario runner (multi-hour phased workloads, per-system telemetry, cross-system plots). #166 adds the adapters we compare against, including **pgq** as the rotation-and-TRUNCATE reference line. Work here starts once #165 ships and at least the pgq adapter from #166 is available; #165 is being implemented first.

## Design-space split

The design space for PG-backed queues splits on hot-path mutation:

| Storage model | Hot path | Vulnerable to MVCC death spiral? | Example |
|---|---|---|---|
| Rotation + TRUNCATE | INSERT only into one of N rotating event tables; whole inactive tables are TRUNCATEd once all consumers have passed them | No on the event-stream path. TRUNCATE is an O(1) metadata op that does not produce dead tuples. A small retry/metadata table still uses UPDATE/DELETE but stays small. | [pgq (Skytools)](https://github.com/pgq/pgq) |
| Per-row state machine | INSERT, UPDATE through states (`available`, `running`, `completed`), DELETE at retention | Yes | Awa (0.5.x), River, Oban, pg-boss, Graphile Worker, pgmq (mostly) |

pgq specifics, from [`structure/tables.sql`](https://github.com/pgq/pgq/blob/master/structure/tables.sql):

[...]

PR #316, opened 2026-06-03
perf(maintenance): exponential prune backoff + branch-tracker hysteresis (#169)

hardbyte:
## Summary

P1 from the recent long_horizon post-mortem on #169. The bench showed Awa's completion rate dropping from 800/s to 392/s during the second pinned-MVCC-snapshot phase with the lease maintenance loop firing every 50ms against a child it couldn't reclaim — every tick attempting `LOCK TABLE ... ACCESS EXCLUSIVE` (50ms timeout) followed by a `SELECT count(*)` that walked dead tuples. The rotate_lease branch also flapped between on-time and delayed at the ~tick-interval boundary, emitting paired warning/recovery events on every flip.

Two coordinated changes in `awa-worker/src/maintenance.rs`:

### 1. `PruneBackoffTracker` — exponential backoff per rotate branch
- When prune returns `SkippedActive` or `Blocked`: `skip_remaining = 1 << level`; bump level (cap at 5 → 32 ticks).
- `Pruned` resets to no backoff (we're healthy again).
- `Noop` is neutral (nothing to prune is not a failure).
- At the 50ms default `lease_rotate_interval`, a pinned snapshot quiets prune attempts to ~1.6s cadence at the cap, dropping promptly back to 50ms once the snapshot is released.
- Rotate step itself (cheap cursor advance) is unaffected; only prune is gated.

### 2. `MaintenanceBranchTracker` hysteresis (K=3)
- Track `consecutive_overrun` / `consecutive_ontime` counters per branch.
- Require K=3 observations in a row before flipping `is_delayed` either direction.
- A single ~tick-boundary jitter no longer fires the warning/counter; a real sustained overrun still does, just one `tick_interval * (K-1)` later.
- Existing #242 contract preserved: still one event per overrun episode, still emitted to `awa_maintenance_branch_overrun_total{branch=...}`.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo build --workspace`
- [x] `cargo test -p awa-worker --lib` (36 passing, 14 new for this PR)
  - 7 `PruneBackoffTracker` tests: initial state, exponential doubling, `SkippedActive`/`Blocked`/`Noop` semantics, `Pruned` reset, saturation at `MAX_PRUNE_BACKOFF_LEVEL`, per-branch independence
  - 7 `MaintenanceBranchTracker` hysteresis tests: single-overrun-no-flip, K-consecutive-overruns-flip, intermittent-no-flip, no re-emit while already delayed, K-consecutive-on-time recovery, per-branch independence

## Follow-ups (out of scope)

- B1 from the same audit: skip `leases.heartbeat_at` write in receipts mode + drop `idx_state_hb` + COALESCE compat reads with `attempt_state.heartbeat_at`. Removes the per-heartbeat non-HOT update entirely.
- #290 single-shard worst-case bench against the new counter table.
- The next `long_horizon` re-run on post-PR main is the validation gate for whether this changes the pinned-snapshot cliff observed in the recent bench.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced verbose logging in maintenance operations to decrease noise.
  * Improved efficiency by minimizing frequency of expensive operations during lock contention through intelligent backoff mechanisms.

### DIFF
--- a/awa-worker/src/maintenance.rs
+++ b/awa-worker/src/maintenance.rs
@@ -4,6 +4,8 @@ use crate::storage::RuntimeStorage;
 use awa_model::cron::{
     atomic_enqueue, list_cron_jobs, upsert_cron_job, CronJobRow, CronMissedFirePolicy,
 };
+#[cfg(test)]
+use awa_model::SkipReason;
 use awa_model::{JobRow, JobState, PeriodicJob, PruneOutcome, RotateOutcome};
 use chrono::Utc;
 use croner::Cron;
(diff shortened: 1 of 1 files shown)


hardbyte:
Combined #316 + #317 long_horizon validation has been posted to #169: https://github.com/hardbyte/awa/issues/169#issuecomment-4617798666\n\nShort version: #316 is directionally useful, but the combined branch still fails idle_1 under pinned MVCC (end depth 223,082, completion ~512/s vs 800/s offered). rotate_lease warning storms also still occur under pinning, so this does not close the #169 gate.

PR #317, opened 2026-06-03
perf(queue_storage): drop leases.heartbeat_at write + state_hb index in receipts mode (#169 B1)

hardbyte:
## Summary

P3 from the long_horizon post-mortem on #169. In receipts mode (the only supported shape for the default `awa` schema), heartbeats land in `attempt_state.heartbeat_at` via the receipts upsert. The `UPDATE leases SET heartbeat_at = clock_timestamp()` that also fired on every heartbeat was redundant — and a non-HOT update because `(state, heartbeat_at)` indexed both columns. Every heartbeat on a running job paid 2 dead index entries per partial index per partition.

Builds on #316 (maintenance backoff + hysteresis). Together these are the two biggest hot-path non-HOT writes from the audit on the leases table.

## Four coordinated changes

1. **`heartbeat_batch` / `heartbeat_progress_batch`** skip the `UPDATE leases` in receipts mode. attempt_state upsert is now the only write. Legacy non-receipts custom schemas keep the old behavior unchanged. Function signatures and return semantics preserved.
2. **Three `LeaseJobRow` SELECTs** read `COALESCE(attempt.heartbeat_at, lease.heartbeat_at)` so `JobRow.heartbeat_at` API surface is preserved in both modes. They already `LEFT JOIN attempt_state`, so no new joins.
3. **`rescue_stale_heartbeats`** skips the leases-side scan in receipts mode. The `rescue_stale_receipt_claims_tx` call already handles rescue from `attempt_state.heartbeat_at`; without the leases write, the leases scan would match zero rows but still walk the table/index.
4. **`idx_<schema>_leases_<slot>_state_hb` dropped** from v023's install helper + new v025 migration. Legacy non-receipts custom schemas fall back to a seq-scan of running rows during rescue (30s cadence, LIMIT 500, bounded by worker concurrency).

## Tests

- `migrate_leaves_no_state_hb_index_on_default_schema` — v025 drops the indexes on `awa`.
- `prepare_schema_does_not_create_state_hb_index` — v023 helper doesn't create them on fresh custom-schema prepare.
- All existing fillfactor + maintenance tests continue to pass.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo build --workspace`
- [x] `cargo test -p awa-model` — full set passing (66 unit + 10 + 7 + 6 + 2 + 2 = ~93 integration)
- [x] `cargo test -p awa-worker --lib` — 28 passing
- [x] Smoke: applied migrations to local DB, confirmed all `idx_*_state_hb` indexes removed via `pg_class`

## Follow-ups (out of scope)

- `awa.jobs_compat()` view (v019) still projects `leases.heartbeat_at` rather than COALESCEing — that's a SQL-only consumer surface; the Rust API path is closed by this PR. A v026 to update that view is the cheap follow-up.
- The next `long_horizon` re-run on post-PR main, paired with the bench from #316, is the gate for whether this changes the pinned-MVCC cliff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed heartbeat tracking for receipt-mode operations to use the correct data source.

* **Performance**
  * Removed unused database indexes to improve storage efficiency and query performance.

* **Tests**
  * Added validation test to ensure heartbeat-related indexes are properly removed.

* **Chores**
  * Updated database schema version to reflect optimization changes.

### DIFF
--- a/awa-model/migrations/v023_install_queue_storage_substrate.sql
+++ b/awa-model/migrations/v023_install_queue_storage_substrate.sql
@@ -909,10 +909,14 @@ BEGIN
             p_schema, v_slot, p_schema, format('leases_%s', v_slot)
         );
 
-        EXECUTE format(
-            'CREATE INDEX IF NOT EXISTS idx_%s_leases_%s_state_hb ON %I.%I (state, heartbeat_at)',
-            p_schema, v_slot, p_schema, format('leases_%s', v_slot)
-        );
+        -- #169 B1: idx_state_hb dropped. In receipts mode (the only
+        -- supported shape for the default `awa` schema)
+        -- `leases.heartbeat_at` is never written, so the index would
+        -- be empty; the rescue path reads `attempt_state.heartbeat_at`
+        -- instead. Legacy non-receipts custom schemas fall back to a
+        -- seq-scan of `WHERE state='running'` rows (bounded by live
+        -- worker count, called at 30s cadence) — acceptable. v025
+        -- drops the index from existing 0.6 deployments.
 
         EXECUTE format(
             'CREATE INDEX IF NOT EXISTS idx_%s_leases_%s_state_deadline ON %I.%I (state, deadline_at)',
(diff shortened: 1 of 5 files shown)


hardbyte:
Combined #316 + #317 long_horizon validation has been posted to #169: https://github.com/hardbyte/awa/issues/169#issuecomment-4617798666\n\nShort version: B1 appears to work (lease/attempt heartbeat churn is no longer dominant), but the combined branch still fails idle_1 under pinned MVCC (end depth 223,082, completion ~512/s vs 800/s offered). So #317 is useful, but not sufficient to close #169 by itself or in combination with #316.

PR #327, opened 2026-06-07
Add queue fanout helper

hardbyte:
## Summary

- add a public `QueueFanout` helper in `awa-model`, re-exported by `awa` and `awa-worker`
- add `ClientBuilder::queue_fanout(&fanout, config)` to declare every physical queue in a logical fanout
- reject duplicate physical queue declarations at build time so manual `queue(...)` calls cannot accidentally overlap with a fanout
- expose the same deterministic fanout router in Python as `awa.QueueFanout`
- document logical queue fanout as the preferred throughput lever when the workload can accept partitioned ordering, with clearer guidance for choosing between fanout, `enqueue_shards`, and `claimers`
- extend ADR-026 with the future append-only terminal-count delta rollup design instead of adding a separate terminal-count ADR
- fix the TLA+ trace harness receipt-only seed actions so they preserve `readyTombstones`, addressing the #325 nightly TLA failure shape
- add adversarial Rust and Python tests for fanout ordering, width bounds, invalid routing keys, invalid queue configs, and duplicate Python fanout queue declarations

## Notes

`queue_fanout` applies `QueueConfig` per physical queue. The docs call out that hard-reserved capacity and per-queue rate limits multiply across the fanout unless callers divide them or use `global_max_workers`.

Python `QueueFanout.queue_configs()` names capacity as `max_workers_per_queue` / `min_workers_per_queue` for the same reason. `route_by_key()` returns insert kwargs (`queue` and `ordering_key`) so it can be passed directly to `insert(...)`. Python `enqueue_many_copy()` is still a homogeneous batch API: one queue and one optional ordering key per call, so mixed-key fanout COPY producers should group batches before calling it.

The ADR-026 terminal-count delta section is future-only. This PR does not change terminal-count writes; exact counts and the existing trusted-counter path remain unchanged.

The TLA+ change is trace-harness only. It does not change the base storage spec; it makes the receipt-only trace scaffolding carry the tombstone ledger through unchanged after #323 introduced `readyTombstones`.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `git ls-files "*.so" ":(glob)**/*.so"` returns no tracked shared objects
- `cargo test -p awa-model queue_fanout --lib`
- `cargo test -p awa-worker queue_fanout_declares_each_physical_queue --lib`
- `cargo test -p awa-worker duplicate_queue_declarations_are_rejected --lib`
- `cargo clippy -p awa-model -p awa-worker --all-targets -- -D warnings`
- `DATABASE_URL=postgres://postgres:test@localhost:15432/awa_test cargo test --workspace`
- `PYO3_PYTHON=/home/brian/dev/awa/awa-python/.venv/bin/python cargo clippy --all-targets -- -D warnings` from `awa-python/`
- `PYO3_PYTHON=/home/brian/dev/awa/awa-python/.venv/bin/python .venv/bin/maturin develop` from `awa-python/`
- `DATABASE_URL=postgres://postgres:test@localhost:15432/awa_test .venv/bin/pytest tests/test_queue_fanout.py tests/test_start_config.py::test_queue_fanout_configs_expand_and_dispatch tests/test_start_config.py::test_duplicate_fanout_queue_config_is_rejected -q` from `awa-python/`
- `DATABASE_URL=postgres://postgres:test@localhost:15432/awa_test .venv/bin/pytest tests/ -q -m "not chaos"` from `awa-python/`
- TLA+ trace harness: snooze, receipt rescue, lost-claim-advance, running cancel, receipt-only cancel, callback wait/resume, DLQ retry, DLQ purge, and broken-trace rejection. The receipt-rescue trace positive witness exits non-zero after completing the intended trace, which is the expected acceptance signal for that witness.
- TLA+ storage gate: `AwaSegmentedStorage`, `AwaSegmentedStorageInterleavings`, safe race configs, `AwaStorageLockOrder`, `AwaDeadTupleContract`, and expected-failure race/deadlock witnesses


hardbyte:
@coderabbitai review

Issue #337, opened 2026-06-10
Failed terminal rows are prunable with no retention floor; admin retry/DLQ-move silently return empty


hardbyte:
## Summary

Queue-ring prune folds **failed** terminal rows into `queue_terminal_rollups` and deletes them from `done_entries` as soon as their slot goes cold — there is no retention floor distinguishing failed rows from completed ones. Once that happens, every admin surface that operates on failed history silently finds nothing:

- `admin::retry_failed_by_kind` / `retry_failed_by_queue` scan `{schema}.done_entries WHERE state = 'failed'` → return `Ok([])`
- `dlq::move_failed_to_dlq` → returns `None`
- discard/purge variants likewise

No error, no warning — the operator's retry just does nothing.

## Reproduction (deterministic)

Found while parallelizing `queue_storage_runtime_test` (#336): with the test client's maintenance leader running at a 1s queue-rotation cadence and `queue_slot_count: 4`, holding the client alive for 30s after two unique-keyed jobs reached `Failed` reproduces it 100%:

- `done_entries`: 0 rows
- `queue_terminal_rollups.pruned_completed_count = 2` ← both rows were **failed**, not completed
- `admin::retry_failed_by_kind(...)` → `Ok([])` where the test expects `Err(UniqueConflict)`

At production cadences the window is minutes-to-hours instead of seconds, but the cliff is the same shape: whether an operator can retry a failed job depends on whether maintenance got there first, and the failure mode is silent.

## Questions to resolve

1. **Should failed rows have a retention floor?** Completed rows aging into rollups quickly is the ADR-026 design working as intended. Failed rows are an operator inbox when `dlq_enabled = false` (the default) — pruning them out of retry reach without any floor effectively makes the DLQ mandatory for any workload that wants reliable manual retry.
2. **Should admin operations distinguish "nothing matched" from "matched rows are pruned"?** `queue_terminal_rollups` knows terminal history existed for the `(queue, priority)`; `retry_failed_by_kind` returning `Ok([])` when the rollup shows pruned terminal counts could instead surface "N terminal rows for this queue have been pruned past retention".
3. **Naming:** failed rows are folded into `pruned_completed_count`. Post-#306 (`QueueCounts.completed → terminal`) the rollup column name carries the same misreading.

## Where the test suite stands

[...]

PR #352, opened 2026-06-14
Compact receipt-backed completions

hardbyte:
## Summary

Make successful short-job receipt completions write **compact, append-only batches** instead of one `done_entries` row per job. On the receipt fast path a completed batch now records:

- one `receipt_completion_batches` row per completion batch (terminal facts as arrays, expanded through retained `ready_entries`),
- one `lease_claim_closure_batches` row per claim as durable close evidence,
- **no** `done_entries` row and **no** per-job terminal-count delta.

`{schema}.terminal_jobs` remains the public terminal read surface and transparently unions `done_entries` with the expanded compact batches, so admin reads, counts, retries, DLQ moves, and discard are unchanged. Direct SQL against `done_entries` should no longer assume every completed job is physically stored there.

This removes the dominant completion-path write amplification (per-job terminal row + count-delta + receipt delete) for the common case — short jobs that complete from the receipt fast path — while keeping every durability and safety guarantee of the queue-storage engine.

## What changed

**Completion plane**
- Successful compact completion appends `receipt_completion_batches` + `lease_claim_closure_batches` and deletes only `attempt_state`. Exact terminal counts are read directly from the retained batch ledger; terminal-count deltas remain reserved for `done_entries` terminal mutations.
- SQL-compatibility deletes of a compact completed job append a `receipt_completion_tombstones` row rather than mutating the batch.
- The slow / wide / non-success paths are unchanged in shape.

**Claim plane**
- Zero-deadline claims are stored compactly in `lease_claim_batches` (arrays + a derived `receipt_ranges` multirange); deadline-backed claims stay row-local in `lease_claims` so the deadline-rescue cursor stays index-backed.
- Claim routes through committed `ready_segments` lane-range metadata before validating `ready_entries`, then runs the attempt-spent battery. The per-row attempt-ledger probe is dropped on the live-head fast path (the head row carries the full ledger probe, and the recovery/spent-prefix path retains it), avoiding a retained-history scan while preserving the lost-post-commit-cursor-advance recovery.
- A compact claim's durable closure is always written to `lease_claim_closure_batches` on **every** closing path (successful completion, non-success, cancellation, rescue, terminal-delete), carrying `ready_slot`/`ready_generation`. A compact claim has no `lease_claims` row, so an explicit `lease_claim_closures` row would be invisible to the queue-ring prune count proof; deadline-backed row-local claims continue to write explicit closures.

**Prune**
- Queue-ring prune truncates the new `receipt_completion_batches`, `receipt_completion_tombstones`, and terminal-delta children alongside the ready/done children, and the count proof balances `lease_claims` + `lease_claim_batches.claimed_count` against explicit closures + `lease_claim_closure_batches.closed_count`.

## Guarantees (unchanged)

- Transactional enqueue, at-least-once delivery, and `(job_id, run_lease)` stale-writer protection are preserved. The authoritative record for a claim is the row-local claim or compact claim-batch item plus durable closure evidence; every spent-evidence reader anti-joins **both** closure ledgers.
- A queue/claim segment cannot truncate while live claims remain; rescue is the gating step. Compact and row-local closures are both partition-truncated, append-only.
- The dead-tuple discipline (ADR-019/023/026) holds: the new ledgers are partition-truncated, not row-vacuumed.

## Migrations

Additive only: `v036` (compact receipt completions), `v037` (ready segments), `v038` (compact claim batches). `CURRENT_VERSION = 38`. Fresh installs (`v023` substrate) and rolling upgrades converge on the same schema; existing schemas are backfilled by per-substrate `DO`-block migrations.

## Formal models

`AwaSegmentedStorage`, `AwaDeadTupleContract`, and `AwaStorageLockOrder` cover the compact completion/claim/closure lifecycle, the dead-tuple kinds of the new ledgers, and the prune lock order. `MAPPING.md` and ADR-019/023/026 are updated to the as-is contract.

## Validation

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo build --workspace`
- `cargo test -p awa --test queue_storage_runtime_test` (full suite, incl. compact-path completion/rescue/cancel/prune and claim-dedup recovery)
- `cargo test -p awa --test migration_test`
- TLC: `AwaSegmentedStorage` (+ interleavings cfg), `AwaDeadTupleContract`, `AwaStorageLockOrder`, `AwaStorageTransition`, `AwaShardedPrune`

## Notes

- Completion sharding: queue storage uses a single completion flusher by default and four for very large single-runtime worker pools (`>= 512` worker capacity), where one serialized flusher cannot keep permits full. Tunable via `AWA_COMPLETION_SHARDS` / `AWA_COMPLETION_BATCH_SIZE` / `AWA_COMPLETION_FLUSH_MS`; extra flushers can multiply fleet-wide contention in multi-process deployments, so the default stays at one.
- No generated `.so`/shared-library artifacts are tracked by git.

## Summary by CodeRabbit

* **New Features**
  * Added compact receipt claim and completion batches for more efficient terminal history storage.
  * Introduced unified `terminal_jobs` view for reading job completion status across compact and physical records.

* **Bug Fixes**
  * Resolved visibility issues where compact claim closures weren't consistently recorded across all exit paths.

* **Chores**
  * Expanded queue-storage schema readiness validation.
  * Updated workflow timeouts and completion batching defaults.
  * Enhanced database migration infrastructure.
  * Strengthened test coverage for receipt-backed operations.


hardbyte:
## Code Review — Compact receipt-backed completions

Reviewed against base `codex/maintenance-prune-nowait` (`git diff origin/codex/maintenance-prune-nowait...HEAD`, 2197 insertions / 27 files). Findings below were each adversarially verified against the actual SQL/Rust before inclusion; B1 was reproduced end-to-end on a local v036 substrate.

### Verdict

**Needs rework — not mergeable yet.** The design is structurally sound: the `terminal_jobs` view expansion is column-faithful, the v036/v023 migration is additive-only, and the TLA+ lock-order/prune edits hold against the SQL. But the fast path moves the durable "this attempt finished" evidence from `done_entries` to `receipt_completion_batches`, and not every reader of the old evidence was taught about the new one — which is where the blockers live. CI is also red on the only failing check ("Telemetry OTLP validation"), for the same class of gap.

---

### Blockers

#### B1 — Claim idempotency battery never consults `receipt_completion_batches` → duplicate execution of a completed job

`claim_ready_runtime`'s attempt-spent probes (`awa-model/migrations/v023_install_queue_storage_substrate.sql:2089-2176` head probe, `:2203-2252` and `:2328-2396` both `selected_with_spent` CTEs) check `ready_tombstones, lease_claims, lease_claim_closures, leases, done_entries, deferred_jobs, dlq_entries, ready_entries` at `run_lease+1` — but **never `receipt_completion_batches`** (0 occurrences in the installed function via `pg_get_functiondef`, vs 3 for `done_entries`).

The successful fast path (`complete_receipt_runtime_batch_fast`, `awa-model/src/queue_storage.rs:5459`, batch insert `:5582`) writes only a `receipt_completion_batches` row plus a terminal-count delta — no `done_entries`, no `lease_claim_closures`, and it leaves the `lease_claims` row in place. That `lease_claims` row lives in the **claim ring**, pruned independently of the **queue ring**. `prune_oldest_claims` (`:12835`) treats the claim as closed via `receipt_closed_evidence_sql` (`:489-518`), which *does* count the batch, so it truncates `lease_claims_<slot>`/`lease_claim_closures_<slot>`. Meanwhile the queue-prune pending guard (`:12290-12320`) retains the ready row while the best-effort post-commit `advance_claim_cursors` (`:3378-3406`) is lost. Net: the only spent-evidence the claim guard *can* see is gone; the only evidence that *remains* (the batch) is invisible to it.

**Concrete interleaving (reproduced live):** claim job 9001 (`run_lease 1`) → fast-complete (batch only; `done_rows=0 closure_rows=0`) → crash before cursor advance → claim-ring prune truncates `lease_claims` → queue-prune skips the ready row (cursor lag) → `claim_ready_runtime` re-claims job 9001 at `run_lease 1` (same idempotency key = same attempt, executed twice). Control (leave `lease_claims` in place) returns 0 rows, isolating the cause to the missing batch probe. The base branch wrote `done_entries` on this path, which the head-spent probe *does* check — so base recovered correctly and this PR regresses it.

> ⚠️ The existing test `test_queue_storage_receipt_claim_dedupes_when_post_commit_cursor_advance_is_lost` (`awa/tests/queue_storage_runtime_test.rs:5245-5294`) **masks** this — it seeds a `done_entries` row a real compact completion never writes, so the `done_entries` probe matches and the bug is hidden. It must be reworked, not relied on.

**Fix:** add a `receipt_completion_batches` EXISTS probe to all three attempt-spent sites, mirroring the receipt-batch arm already in `receipt_closed_evidence_sql` (`:499-518`) — ideally by factoring a single shared evidence helper so the claim battery and `receipt_closed_evidence_sql` stay in lockstep. Alternatively, gate `prune_oldest_claims` so a claim partition can't be truncated until the lane's claim cursor has advanced past the completed lane. Add a regression test that fast-completes, drops only `lease_claims`/`closures` (leaving the batch + ready row), and asserts `claim_ready_runtime` returns no rows.

#### B2 — Red CI: "Telemetry OTLP validation" fails deterministically

The only failing check, caused by M1 below (`test_collector_death_does_not_block_job_processing` times out, panic at `telemetry_test.rs:682`, exit 101). Must be green to merge.

---

### Major

#### M1 — Telemetry test `running` detection misses compact-batch completions (double-counts completed jobs)

`awa/tests/telemetry_test.rs:262-266` (`queue_state_breakdown`) and `:188-192` (`queue_job_count`) count each `lease_claims` row as `running` unless one of five `NOT EXISTS` arms fires (`lease_claim_closures, leases, deferred_jobs, done_entries, dlq_entries`). The fast path writes none of those, so completed jobs stay `running` forever — and the terminal arm (correctly updated to read `terminal_jobs`) *also* counts them, so each completed job is counted twice. Reproduced: `Timed out waiting for queue ... to drain; live_jobs=N, breakdown=[("running", N)]`.

Production closure-idempotence uses `receipt_closed_evidence_sql` and is unaffected (no data loss) — this is a test/CI defect, but it is the red check. **Fix:** add a sixth `NOT EXISTS` arm to both running-detection blocks excluding `lease_claims` rows present in `receipt_completion_batches`, mirroring `receipt_closed_evidence_sql` — better, have the test reuse that helper rather than maintain a parallel exclusion list that has now drifted.

#### M2 — Receipt-path terminal-count deltas use constant `counter_bucket=0` while rebuild re-buckets by `job_id%256` → decrements clamped away, counter stuck high

The fast path writes its `+1` delta at hardcoded `counter_bucket=0` (`awa-model/src/queue_storage.rs:5680`), and the v036 compat-delete writes its `-1` at `0` (`awa-model/migrations/v036_compact_receipt_completions.sql:524`). But `rebuild_terminal_counters` (`queue_storage.rs:11737-11744`) re-derives each row's bucket as `mod(job_id,256)`, and the rollup fold (`:12085`) does per-bucket `GREATEST(0, live + delta)` with INSERT gated on `delta > 0`. Base used `mod(job_id,256)` here; the PR changed it to a constant.

**Interleaving:** (1) fast-complete job 1000 → `+1@bucket0`; (2) operator runs `rebuild_terminal_counters` (the documented drift-recovery op) → places job at `bucket232`, wipes deltas; (3) compat-delete of 1000 → `-1@bucket0`; (4) rollup folds `bucket0`: UPDATE matches no live row (it's at 232), INSERT skipped (`delta<0`) — the `-1` is truncated away. Net: counter permanently one too high, violating the documented invariant `SUM(live)+SUM(deltas) == count(*) FROM terminal_jobs` (`:4053-4054`), in the very counter `rebuild` exists to make billing-grade. **Fix:** bucket the receipt-path deltas by `job_id` (`mod(mod(job_id,256)+256,256)::smallint`) at both sites and GROUP BY accordingly.

---

### Minor

- **`correctness/storage/MAPPING.md:28`** (echoed at `:93`) — prose claims the compact success path writes a `'completed'` closure row; it doesn't (closure is derived from the batch via `receipt_closed_evidence_sql`). Contradicts the doc's own `FastComplete` row at `:40`. The TLA+ specs are fine — only the prose. A literal `'completed'` closure row is written only on the slow fallback (`complete_runtime_batch_slow_in_tx`, `:5843-5859`).
- **`awa-model/src/queue_storage.rs:219`** — public `QueueCounts::terminal` doc says "`count(*) FROM done_entries` semantics"; now spans both tables. The inline invariant at `:4054` already says `terminal_jobs`; four sibling sites were updated, this public field doc was missed.
- **`awa-model/migrations/v036_compact_receipt_completions.sql:507-529`** — the receipt-batch delta INSERT passes a 6th arg (`p_id`) in `USING` that's never referenced (harmless but misleading). Becomes live if M2 is fixed by bucketing on `job_id` — coordinate the two.
- **`CHANGELOG.md` `[Unreleased]`** — no entry. This adds migration v036, two public `BusyCounts` fields (`queue_receipt_completion_batches`, `queue_receipt_completion_tombstones`), two metrics, and changes the meaning of direct SQL against `done_entries` for completed jobs. Comparable storage changes (#337, #348) got entries; add before merge and call out the `done_entries` semantic change explicitly.

---

### What's solid (verified — no need to re-litigate)

- **`terminal_jobs` view expansion** — column-by-column faithful: `done_entries UNION ALL (receipt_completion_batches` expanded via `CROSS JOIN LATERAL UNNEST(job_ids, run_leases) WITH ORDINALITY` against retained `ready_entries`); state/run_lease/job_id alignment correct (`v023:1356-1431`).
- **Migration discipline** — v036 net-new and additive-only; v023 substrate changes additive (no drop/rename/retype/tighten).
- **TLA+ fidelity** — `AwaSegmentedStorage.FastComplete/StatefulComplete` adding `<<j,runLease>>` to `claimClosed` is a faithful abstraction (the batch is durable closure evidence); `AwaDeadTupleContract.CompleteReceiptsTx`, the prune truncate/lock lists, `DeleteCompactReceiptTerminalCompatTx`, the `PruneReadySegment` cursor-guard change, and all `AwaStorageLockOrder` plan edits match the SQL.
- **`receipt_closed_evidence_sql` (`:489-537`)** is the correct, complete evidence helper — it *does* consult the batch. The blockers/majors are precisely the readers that should have reused it but didn't.
- **Refuted, for awareness:** the Python chaos test and bench/ADR-artifact docs were flagged and cleared — rescued completions route through the *slow* path (materialized lease → writes `done_entries`), so those readers are correctly unaffected. The bug surfaces specifically where the fast/receipt path is the completion route (the default for plain receipt-eligible jobs, *not* for deadline-rescued re-claims).

---

### Open questions

1. **Single source of truth for closure evidence** — will you factor one shared "is this attempt's terminal evidence durable?" helper used by `claim_ready_runtime`'s attempt-spent battery, `receipt_closed_evidence_sql`, and the telemetry test SQL? Three hand-maintained copies already diverged (B1, M1). Any reason the claim battery can't reuse `receipt_closed_evidence_sql` directly?
2. **Prune ordering invariant** — is there meant to be an invariant that claim-ring prune must not run before the queue cursor has advanced past a completed lane? It isn't enforced today; if not intended, B1's interleaving is reachable by design.
3. **Bucketing rationale (M2)** — was the constant `counter_bucket=0` deliberate (collapse contention on one bucket) or a regression from base's `mod(job_id,256)`? If deliberate, how should `rebuild_terminal_counters` reconcile, given it re-buckets by `job_id`?
4. **Draft scope** — are the telemetry-test fix and CHANGELOG entry intentionally deferred to ready-for-review, or missed?

hardbyte:
## Code Review (round 2) — Compact receipt-backed completions

Re-reviewed at HEAD `8d646734` against base `codex/maintenance-prune-nowait`. The PR grew from ~2.2k to ~4.8k insertions across 43 files since round 1 (new `v037_ready_segments.sql` + ready-segment claim routing, prune-lock relaxation). Each finding below was adversarially verified against current code; the round-1 fixes were re-checked, not taken on faith.

### Verdict

**Mergeable after one more round.** The core compact-completion mechanism is sound, and the four round-1 blockers/majors (B1, B2, M1, M2) are **genuinely closed** — CI telemetry is green again. But the prune-lock relaxation introduced this round opens a real (if narrow) data-loss TOCTOU window (M-A), and the `reset()` edit reintroduces the B1 silent-skip class for any standalone caller (M-B). Both should be fixed before this leaves draft. Two CI jobs (*Rust tests*, *Python build + test*) were still pending at review time — confirm green.

No blockers confirmed this round.

---

### Major

#### M-A. Prune-lock relaxation drops lock-first ordering and never re-checks the skip gates after `ACCESS EXCLUSIVE` → TOCTOU data-loss window
**`awa-model/src/queue_storage.rs:12756–12912` (`prune_oldest`) and `13486–13580` (`prune_oldest_claims`)**

This round moved the prune skip-gates to run *before* `LOCK TABLE … IN ACCESS EXCLUSIVE MODE` and dropped the base branch's `NOWAIT` (line 12901 / 13547). The gates `return` on success before the lock (`active_leases` 12780, `pending` 12821, `unclosed_claim_refs` 12889), and there is **no re-check of `pending` between the lock and the `TRUNCATE`** at 13125. The claim-ring variant has the same shape: the `open_claims` gate runs at 13517–13542 before the lock, and the only post-lock recheck (13564–13580) re-verifies `current_slot != slot` — **not `open_claims`**.

The invariant the new code relies on ("the ring lock seals the slot, so an in-flight producer/claimer must still hold a conflicting child-table lock") does not hold: producers read the ring with a plain unlocked `SELECT` (`current_queue_ring`, `queue_storage.rs:3310`, no `FOR UPDATE`) and never take the ring lock. `rotate` only bumps the *next* slot's generation (12016+), so a producer's committed rows carry the same `(slot, gen)` the prune target query matches (12728–12736).

**Concrete interleaving** (near-startup, one rotation; slot 0 seeded gen 0):
1. Producer T1 reads ring → slot=0, gen=0; begins `INSERT` into `ready_entries_0` (RowExclusive), uncommitted.
2. `rotate` advances 0→1; slot 0 keeps gen 0; T1 not blocked (took no ring lock).
3. `prune_oldest` locks `queue_ring_state` FOR UPDATE (current=1), target=(0, gen 0); `pending`/`active_leases`/`unclosed` gates all pass — T1's INSERT is uncommitted, hence MVCC-invisible.
4. `LOCK TABLE ready_entries_0 … ACCESS EXCLUSIVE` blocks on T1's RowExclusive (`lock_timeout=10ms`, `PRUNE_LOCK_TIMEOUT`).
5. T1 commits within 10ms; its slot-0/gen-0 rows are now committed and visible.
6. Prune unblocks, `TRUNCATE`s `ready_entries_0` (13125) + deletes its `ready_segments` row (13138) with **no `pending` re-check** → a committed enqueued job is silently destroyed.

On the base branch this window was closed: the lock was acquired *first* with `NOWAIT`, so either T1 was mid-insert and the prune got `Blocked`, or T1 had committed/released and the post-lock `pending` gate saw the rows and skipped. The claim variant orphans a running job's closure evidence the same way.

**Why major, not blocker:** the 10ms `lock_timeout` bounds the window, and the ring must be positioned so the producer's slot is the oldest non-current target — narrow and timing-dependent. But the outcome is silent destruction of a committed job, which meets the data-loss bar.

**On the TLA+ change:** `AwaStorageLockOrder.tla` re-models lock-*acquisition order* only and explicitly excludes MVCC/snapshot isolation. A passing model that abstracts away the visibility semantics central to this TOCTOU does **not** cover this race — please don't cite it as evidence the relaxation is safe.

**Fix direction:** either (a) restore lock-first ordering — take `LOCK TABLE … ACCESS EXCLUSIVE` (keep `NOWAIT` or the 10ms timeout so a contending producer yields `Blocked`) *before* the gates; or (b) re-run the relevant gate(s) *after* acquiring the lock and before `TRUNCATE` — re-check `pending` (and ideally `active_leases`/`unclosed_claim_refs`) in `prune_oldest`, and `open_claims` (not just `current_slot`) in `prune_oldest_claims`. Once the lock is held the gate result is stable, so one cheap post-lock re-check closes the window.

#### M-B. `reset()` omits the three new compact tables from its `TRUNCATE`, reintroducing the B1 silent-skip for reused job_ids
**`awa-model/src/queue_storage.rs:2897–2917` (TRUNCATE list) + `:2924` (`job_id_seq RESTART WITH 1`)**

`reset()`'s `TRUNCATE` was updated this PR to add `ready_segments` (2900) but **omits `receipt_completion_batches`, `receipt_completion_tombstones`, and `lease_claim_closure_batches`** — the three tables this PR introduces. Because `receipt_completion_batches` survives reset while `job_id_seq` restarts at 1 (2924) and `queue_ring_state` generation resets to 0, a stale batch row collides with a reused `job_id` at the same `(ready_slot, ready_generation, queue, priority, enqueue_shard)` lane: the head probe at `v023:2491–2512` matches `job_ids @> [v_head_job_id]` plus a member `run_lease = v_head_run_lease + 1`, sets `v_head_attempt_spent`, advances the cursor past the new job, and it is **silently never executed** — the exact B1 class this PR fixes, reintroduced via `reset()`.

**Why major, not blocker:** all three current callers (`awa-python/src/client.rs:679`, `:2582`; test helper `recreate_store_schema` at `queue_storage_runtime_test.rs:303`) precede `reset()` with `DROP SCHEMA … CASCADE`, which physically empties the compact tables — so **no live data-loss path in shipped code today**. The method is safe only by coincidence of caller ordering; it violates its own clean-slate contract for any future/external caller relying on `reset()` alone.

**Fix direction:** add the three tables to the TRUNCATE list (mirroring how `ready_segments` was added — they're partitioned parents, so truncate cascades to children). Consider also resetting `receipt_id`/`queue_enqueue_heads`/`queue_claim_heads` sequences for full clean-slate consistency with `job_id_seq RESTART`. Add a regression test that completes via the compact path, calls `reset()` **standalone (no DROP SCHEMA)**, re-enqueues reusing `job_id=1`, and asserts it is executed rather than skipped.

---

### Minor

#### m-1. Stale `queue_storage.rs:NNNN` citations in MAPPING.md and AwaStorageLockOrder.tla
**`correctness/storage/MAPPING.md` and `correctness/storage/AwaStorageLockOrder.tla`** (multiple `:NNNN` refs)

`queue_storage.rs` grew to ~13,634 lines; model→code citations drifted 500–1180 lines and now land on unrelated lines: `prune_oldest` cited `:11689` → actual 12708; `prune_oldest_claims` `:12260` → 13441; `prune_oldest_leases` `:12021` → 13185; `rotate_claims` `:12154` → 13326; `.tla close_receipt_tx :6517` → 7033; `rescue_stale_receipt_claims_tx :8195` → 8894; `ensure_running_leases_from_receipts_tx :7574` → 8241; `cancel_job_tx :6568` → 7184. The prune/rotate MAPPING rows were rewritten this round while reusing the stale numbers. These docs are the auditable bridge the formal-fidelity review depends on. **Fix:** replace `:NNNN` with function-name anchors (e.g. `QueueStorage::prune_oldest_claims`), or regenerate against HEAD.

---

### What's SOLID (verified — do not re-litigate)

- **B1 — FIXED.** All gating spent-evidence probes now consult `receipt_completion_batches`: the single-head probe (`v023:2491–2513`) and the recovery CTE `selected_with_spent` (`v023:2612–2630`) match the lane head against `job_ids @>` + an `unnest(job_ids, run_leases)` pair where `run_lease = v_head_run_lease + 1` — correct "next attempt already completed" semantics, with cardinality-checked positional arrays. The masking test was genuinely reworked (`queue_storage_runtime_test.rs:5598–5694`): zero `done_entries`, real compact completion, deletes `lease_claim_closure_batches`/`lease_claim_closures`/`lease_claims` leaving `receipt_completion_batches` as sole evidence, asserts no re-emit + cursor advance.
- **B2 — FIXED.** `Telemetry OTLP validation` green on HEAD; the job runs `cargo test -p awa --test telemetry_test -- --ignored` against live Postgres + otel-lgtm, so not a no-op mask.
- **M1 — FIXED.** Both telemetry helpers (`telemetry_test.rs:207–210`, `:287–289`) exclude compact completions from the running arm via `NOT EXISTS (… lease_claim_closure_batches … receipt_ranges @> lc.receipt_id)` + `lc.closed_at IS NULL`, and source terminal state from the `terminal_jobs` view — each compact completion counts exactly once.
- **M2 — FIXED.** All four bucket-producing sites agree on `mod(mod(job_id,256)+256,256)`: fast-path CTE (`queue_storage.rs:5913–5917`), v036 compat-delete INSERT (`v036:412`) and UPDATE (`v036:426`), and `rebuild_terminal_counters` (`queue_storage.rs:12178–12181`). The decrement-clamped-after-rebuild mode can no longer occur.
- **New ready_segments routing — correct.** The claim LATERAL (`v023:2357–2410`) uses `ready_segments` only as a compact index to pick `(ready_slot, ready_generation)` by lane range, then reads `ready_entries` and runs the *same* spent-evidence probes before emitting. It cannot route to a pruned slot (prune deletes the segment row in the same txn as the truncate); a spent head routes into the recovery/cursor-advance path, not re-execution. Column fidelity holds (DDL `v023:407–419` ↔ insert `v023:1938–1957` ↔ Rust `insert_ready_segments_tx`).
- **v037** is a correct additive re-apply/backfill migration; `CURRENT_VERSION=37`.
- **Round-1 doc minors fixed:** MAPPING.md:28 now describes the compact closure-batch write (not a `done_entries 'completed'` row); `QueueCounts::terminal` doc reflects dual storage; the `#352` CHANGELOG entry is accurate including the `done_entries`-semantics caveat.

---

### Open questions
1. **M-A:** Was the lock-first → gate-first inversion intentional (to avoid blocking hot claim reads behind a doomed prune)? If so, why no post-lock re-check? Restore `NOWAIT`/lock-first, or add the post-lock gate re-check?
2. **M-B:** Is `reset()`'s contract "in-place clear to clean slate" or "only ever called after `DROP SCHEMA`"? If the latter, the `DROP SCHEMA` in callers makes the `TRUNCATE`+`RESTART` redundant — which is the source of truth? Either way the three new tables belong in the list.
3. The prune comment at `queue_storage.rs:13491–13495` asserts closure evidence is "monotonic until this same prune path truncates the slot" — true for *existing* claims but not against a *new* open claim committed into a just-sealed slot. Please reconcile with the M-A fix.
4. Confirm *Rust tests* and *Python build + test* go green before un-drafting.

hardbyte:
## 60-minute Awa-only soak after compact receipt completions

Run: `custom-20260614T231041Z-752523`
Command shape: `awa`, 5k/s offered, 512 workers, 30s warmup + 60m clean, local benchmark harness.

Clean phase summary from 720 samples:

| metric | value |
| --- | ---: |
| median enqueue rate | 4,991/s |
| median completion throughput | 4,991/s |
| estimated completed jobs | ~18.0M |
| median queue depth | 250 |
| queue depth p95 / p99 / max | 12,847 / 19,833 / 26,261 |
| median e2e p99 | 427 ms |
| e2e p99 p95 / p99 / max | 4,069 / 4,592 / 5,313 ms |
| WAL delta | 34.0 GB / 235.0M records |
| approx WAL per completed job | 1.89 KB / 13.1 records |

Dead-tuple signal:

- `ready_claim_attempt_batches_*`: final summed `n_dead_tup = 0` across all 16 partitions.
- `receipt_completion_batches_*`: final summed `n_dead_tup = 0` across all 16 partitions.
- `queue_terminal_count_deltas_*`, `ready_entries_*`, `done_entries_*`: zero dead tuples in sampled hot partition families.
- Remaining visible dead tuples were mostly tiny mutable metadata tables and `ready_segments` churn, not the compact hot batch tables.

Instrumentation captured during waves:

- `instrumentation/manual/late_wave_wait_summary_awa_bench.txt`
- `instrumentation/manual/final_wave_wait_summary_awa_bench.txt`
- `instrumentation/manual/*_pg_stat_statements.txt`
- `instrumentation/manual/*_activity_tables_awa_bench.txt`

Named residual mechanism: **bounded-concurrency WAL work with catch-up waves**, not dead-tuple collapse. During the largest captured wave, `pg_stat_activity` showed 3 active backends, no lock waiters, and no WAL wait sampled at that instant; the compact batch tables still had zero dead tuples. The long-run median sustains the offered 5k/s, but short intervals dip to ~3.1-3.3k/s completion and then catch up above offered rate, producing queue-depth and p99-latency waves. `pg_stat_statements` still shows the durable claim path, enqueue COPY, and compact completion statements as the dominant SQL/WAL budget.

hardbyte:
@coderabbitai review

hardbyte:
## 60-minute Awa-only soak after `c6f4b82`

Run: `custom-20260615T184208Z-08dc51`

Shape: Awa only, 5k/s offered, 512 workers, Postgres 18.3, `POSTGRES_CPUS=4`, 30s warmup + four 15-minute clean phases. Awa revision was `c6f4b82` (`codex/compact-claim-batches`, pushed as `codex/compact-receipt-completions`); benchmarking harness revision was `6d7a77a`.

### Headline

This PR does what it is meant to do for hot-table tuple churn: the append-only/compact tables stayed effectively free of dead tuples. It does **not** make a single physical queue sustain 5k/s for an hour on this machine/config.

Approximate integrated clean-hour totals from sampled rates:

| phase | enqueued | completed | net backlog | completion median | final depth |
| --- | ---: | ---: | ---: | ---: | ---: |
| `clean_15m_a` | 4.50M | 4.32M | +0.18M | 4,965/s | 181,576 |
| `clean_15m_b` | 4.50M | 4.68M | -0.18M | 4,625/s | 251 |
| `clean_15m_c` | 4.50M | 3.10M | +1.40M | 2,663/s | 1,395,232 |
| `clean_15m_d` | 4.53M | 3.27M | +1.26M | 2,265/s | 2,655,593 |
| **clean total** | **18.02M** | **15.37M** | **+2.65M** | **3,433/s overall median** | **2,655,593** |

The p99 latency samples become unhelpful late in C/D because the sampled latency windows report zeros while depth is still growing. Queue depth/backlog is the reliable signal there.

### Tuple/WAL evidence

- Hot append-only/compact storage dead tuples stayed flat: `ready_entries_*`, `ready_segments_*`, `ready_claim_attempt_batches_*`, `receipt_completion_batches_*`, `receipt_completion_tombstones_*`, `queue_terminal_count_deltas_*`, `done_entries_*`, `lease_claim_batches_*`, and `lease_claim_closure_batches_*` peaked at **1 dead tuple** total in the sampled hot set.
- `queue_terminal_live_counts` remained at zero live/dead tuples in the sampled phases.
- Remaining dead tuples were small control-plane churn: claim/lease/queue ring state and claimer lease rows, with sampled control-plane peaks under ~80 rows. `attempt_state` peaked at 329 in A; `deferred_jobs` peaked at 500 during the late rescue-amplified phase.
- WAL remained steady at about **4.1 GiB per 15-minute clean phase**; total clean-phase WAL delta was about **16.5 GiB**. COPY remains the largest WAL producer, as expected.

### Statement/wait evidence

Late `pg_stat_statements` cumulative totals were dominated by the claim path and compact completion path:

- `SELECT ... FROM awa.claim_ready_runtime(...)`: 2.47M ms total, 90k calls, mean 23.3 ms.
- Inner `claim_ready_runtime` SQL: 2.03M ms total, 64.8k calls, mean 26.0 ms.
- Compact completion statements: several partition-specific statements in the 200-317k ms total range; late means rose to ~30-44 ms per batch statement.
- COPY enqueue: 285k ms total, mean ~2.0 ms per COPY call, and ~11.9 GiB WAL.
- Receipt stale-rescue scan became visible late: 834 calls, 27.7k ms total, mean 33.3 ms, plus several >1s slow-query warnings once backlog had aged.

Claim `EXPLAIN (ANALYZE, BUFFERS)` boundary snapshots:

| phase | rows | execution | shared hits | shared reads | WAL bytes |
| --- | ---: | ---: | ---: | ---: | ---: |
| A | 0 | 27.7 ms | 4,071 | 164 | 718 |
| B | 32 | 47.1 ms | 7,806 | 5 | 30,151 |
| C | 32 | 40.6 ms | 5,847 | 136 | 29,963 |
| D | 0 | 33.0 ms | 4,194 | 76 | 192 |

Wait samples were mostly CPU and WAL sync/write, not blocking:

- CPU: 58-68% of active wait samples by phase.
- `IO:WalSync`: 16-25%.
- `LWLock:WALWrite`: 7-12%.
- No broad lock pile-up and no stuck xmin. Manual snapshots saw only short active transactions.

### Named residual mechanism

I would call the remaining failure mode **single-queue receipt-cycle saturation with rescue amplification**.

The compact append-only storage removes the original dead-tuple collapse. The bottleneck shifts to per-batch claim and completion proof work for one hot physical queue. Once that path falls behind, backlog grows; once claimed receipt jobs age beyond the heartbeat staleness window, maintenance starts sweeping claim batches. Those rescue scans are safety-preserving, but they add more DB work and cause late worker completions to be ignored as already rescued/cancelled. That amplifies the degraded state rather than causing the initial saturation.

Artifacts are in:

`results/custom-20260615T184208Z-08dc51/`

Manual live snapshots are under:

`results/custom-20260615T184208Z-08dc51/instrumentation/manual/slow*_*.txt`

### Implication

This PR is still a correctness/perf improvement: it validates that the hot completion path is no longer a dead-tuple generator. The next throughput work should not be more vacuum tuning. The highest-value follow-ups are:

1. Reduce per-batch claim/completion proof work for compact receipts.
2. Make the receipt heartbeat rescue path avoid expensive false-positive sweeps when the receipt-plane worker is merely completion-backlogged, not crashed.
3. Treat queue fanout/partitioned queues as the way to scale beyond a single physical queue’s coordination budget.

hardbyte:
## Code Review (round 3) — Compact receipt-backed completions

Re-reviewed at HEAD `ac1be16d` against base `codex/maintenance-prune-nowait` (16 new commits since round 2: compact claim batches `v038`, partitioned/cached ready-segment routing, and compact completions no longer writing terminal count deltas). Round-2 findings were re-verified against current code; the new mechanics were reviewed fresh. Each finding below survived an adversarial refutation pass.

### Verdict

**Mergeable after fixes — one more correctness pass, not a rewrite.** Both round-2 majors are genuinely closed with real regression tests (M-A prune TOCTOU, M-B `reset()` truncate gap), and the new delta-free count path is exact. But this round's new claim-batch + rescue interaction reopens the exact cross-ring coupling that `8c33cf2b` ("Decouple claim cursor recovery from claim ring prune") set out to remove: a **rescued compact (zero-deadline) claim permanently fails the queue-prune safety gate**, stalling queue-ring space reclamation until the claim ring independently catches up. That's a real liveness regression and the gating fix for this round.

**Correction to my round-2 comment:** the "red telemetry CI" I flagged as blocker B2 was a *misattribution*. The failing job at that commit was **Rust lint** (`clippy::needless_borrows_for_generic_args` on a `.bind(&[...])` literal in `migration_test.rs`), which never reached the telemetry job — now `.bind([...])` and passing. The telemetry test itself was fine. Apologies for the mislabel; M1's underlying running-detection fix was real and is confirmed below.

No blockers confirmed this round.

---

### Major

#### M3-A (new). Rescued compact claim permanently blocks queue-ring prune — reopens cross-ring coupling
- **Gate:** `awa-model/src/queue_storage.rs:694-703` (`queue_prune_has_unclosed_claim_refs_tx`, `explicit_count` CTE)
- **Rescue writes:** `:9977` (heartbeat rescue) and `:9531` (`close_open_receipt_claim_tx` / deadline rescue)
- **Contrast — normal completion:** `:6097` writes `lease_claim_closure_batches` **with** `ready_slot`/`ready_generation`

The queue-prune safety gate decides reclaimability by:
`count(lease_claims) + sum(lease_claim_batches.claimed_count) == explicit_closures + sum(lease_claim_closure_batches.closed_count)`.
`explicit_count` derives its per-generation filter by **JOINing `lease_claim_closures` to `lease_claims`** on `(claim_slot, job_id, run_lease)` and filtering `lease_claims.ready_slot/ready_generation`.

A compact (zero-deadline) claim has **no `lease_claims` row** — v023's claim CTE splits on `p_deadline_secs`: `<= 0` inserts only into `lease_claim_batches`, `> 0` only into `lease_claims`. Both rescue paths write closure evidence into the **explicit** `lease_claim_closures` table (`claim_slot, job_id, run_lease, outcome, closed_at` — no ready coords), never into `lease_claim_closure_batches`. (`:9977`: `stale_locked = stale_row_locked UNION ALL stale_batch_locked`, so compact batch claims do flow into this explicit insert.)

For a rescued compact claim in a sealed generation:
- `compact_claim_count` = 1 — the `lease_claim_batches` row survives (**no `DELETE FROM lease_claim_batches` exists anywhere**, confirmed by grep)
- `explicit_count` = 0 — no `lease_claims` row to join
- `compact_count` = 0 — rescue wrote no closure-*batch*
- `0 + 1 == 0 + 0` → false → `prune_oldest` returns `SkippedActive{QueueUnclosedClaimRefs}` for that slot, **permanently**, until the claim ring acts.

**Interleaving:** enqueue zero-deadline job (ready 0/gen 0) → claim compact (only `lease_claim_batches_0`; 0 rows in `lease_claims`) → worker crashes → `rescue_stale_heartbeats` writes one explicit `lease_claim_closures` row → rotate queue ring off slot 0 → `prune_oldest(0)` returns `SkippedActive{QueueUnclosedClaimRefs, count:1}` on every attempt; ready/done/receipt_batch/segment children stay un-truncated and `rotate()`'s busy-check refuses to lap onto the slot.

**Why it hides:** the claim-ring gate `claim_prune_has_open_claims_tx` (`:731-771`) counts the closure child **directly** (no `lease_claims` JOIN), so it *does* see the rescue closure (`0+1 == 1+0`), prunes the `lease_claim_batches` partition, after which the queue gate rebalances (`0==0`). So queue reclamation is now gated on independent claim-ring progress — the coupling `8c33cf2b` claimed to remove. Liveness/space-reclamation, not data loss (the job is correctly recorded spent) → major, not blocker.

**Fix direction (either):**
(a) rescue paths write a `lease_claim_closure_batches` row carrying `ready_slot`/`ready_generation` for rescued **compact** claims (mirror the normal completion at `:6097`); or
(b) add a term to `explicit_count` that counts `lease_claim_closures` matching an *unnested* item of a `lease_claim_batches` row for this `ready_slot`/`ready_generation` (join closures to the unnested batch items, not to `lease_claims`).
Regression test: zero-deadline job → compact claim → stale → rescue → rotate queue ring off the slot → assert `prune_oldest` returns `Pruned` **without** first pruning the claim ring. Confirm the fix covers **both** the heartbeat (`:9977`) and deadline (`:9531`) rescue paths.

---

### Minor

#### m3-1. Stale `queue_storage.rs:NNNN` citations remain in `AwaDeadTupleContract.tla` (round-2 m-1 surviving in one file)
`correctness/storage/AwaDeadTupleContract.tla:274, 281, 294, 307, 315, 356, 363`. Round-2 m-1 was correctly fixed in `MAPPING.md` and `AwaStorageLockOrder.tla` (both now function-name anchored, zero `:N` refs) — but `AwaDeadTupleContract.tla` was missed, and its 7 numeric citations drifted ~900–1900 lines this round. Now actively misleading: e.g. `close_receipt_tx` cited `:6517`, actual `:7411`; the `:10081` ref lands on a *different* function (`rescue_expired_receipt_deadlines_tx`). The function-name anchors are still correct. Fix: strip the `: queue_storage.rs:NNNN` suffixes to match the other two specs.

#### m3-2. `storage_abort` emptiness gate omits `lease_claim_batches`
`awa-model/migrations/v037_ready_segments.sql:125-143` (the `storage_abort` emptiness sum; not updated by v038). The mixed-transition abort guard enumerates queue-storage tables to prove emptiness — `lease_claims`, `lease_claim_closures`, `lease_claim_closure_batches`, `ready_claim_attempt_batches` — but **not** `lease_claim_batches`, which uniquely holds in-flight zero-deadline compact claims. The guard is meant to be exhaustive (`docs/migrations.md:198`). Minor, not major, for a specific reason: for any *open* compact claim, a *gated* co-resident table is also non-empty (`ready_entries`, append-only `ready_claim_attempt_batches`), so the abort is rejected in the common case and live compact jobs aren't abandoned — but it's a genuine exhaustiveness/contract gap. Fix: ship a later migration redefining `storage_abort` with `+ (SELECT count(*) FROM %1$I.lease_claim_batches)` (don't edit v037 if already applied).

---

### What's solid (verified — do not re-litigate)

- **M-A (prune TOCTOU) genuinely closed.** Queue prune rechecks all three gates — `active_leases` (`:13779`), `pending` (`:13790`), `unclosed_claim_refs` (`:13801`) — **after** `LOCK TABLE … ACCESS EXCLUSIVE` (`:13762`) and **before** TRUNCATE. Claim prune rechecks `current_slot` **and** `open_claims` (`:14448`) post-lock. Load-bearing on READ COMMITTED (no isolation override), so committed-during-lock-wait rows are visible. Two regression tests poll `pg_stat_activity` to confirm the prune actually blocks on the child lock, then assert `SkippedActive` + row survival.
- **M-B (reset truncate) genuinely closed.** `reset()` (`:3113-3137`) now truncates all compact tables (`receipt_completion_batches`, `receipt_completion_tombstones`, `lease_claim_batches`, `lease_claim_closure_batches`, `ready_segments`, `ready_claim_attempt_batches`); standalone regression `test_queue_storage_reset_truncates_compact_receipt_evidence` exercises the exact B1 reuse-after-restart scenario.
- **New delta-free count path is exact** (supersedes round-1 M2's basis). Compact completions write no `queue_terminal_count_deltas`; `queue_counts_exact` (`:6874-6913`) instead adds `SUM(receipt_completion_batches.completed_count) − count(receipt_completion_tombstones)`, mapping exactly to the two `UNION ALL` arms of `terminal_jobs`. Done vs batch completions are mutually exclusive; compat-delete subtracts via `−1` delta (done path) **or** tombstone (batch path), never both. `rebuild_terminal_counters` rebuilds only from `done_entries` — correct, since batches are counted directly.
- **B1 still intact under the new claim-batch proof.** A rescued job *is* recorded spent (closure row exists; B1 probes read `lease_claim_closures` directly). M3-A is prune liveness, not a reopened silent-skip.
- **CI lint fixed** (see Verdict correction).

---

### Open questions
1. **M3-A scope:** confirm the chosen fix covers the *deadline* rescue path (`:9531`) as well as heartbeat rescue (`:9977`) — the gate analysis applies identically to both.
2. **Bounded-proof shortcut:** `queue_prune_has_unclosed_claim_refs_tx` returns skip on *any* mismatch and avoids the exact anti-join under a hot MVCC horizon (`:725-729`). With M3-A the mismatch is now *permanent* for rescued compact claims, not transient. Does the fix make the equation balance, or does the shortcut still leave a permanently-skipped slot? The regression test should assert eventual `Pruned`, not merely "no panic".
3. **`storage_abort` exhaustiveness:** add a set-diff parity test asserting the abort emptiness enumeration covers *every* table created by `install_queue_storage_substrate`, so future compact tables can't silently drop out of the guard again.

hardbyte:
<html><head></head><body><h2>Benchmark update after <code>fba2d86</code></h2><p><code>fba2d86</code> avoids claim-attempt ledger probes on live-head claims.</p><p>The previous no-env 4 × 15m, 5k/s run, <code>custom-20260615T225956Z-896835</code>, still failed in Phase B:</p><ul><li><p>Depth climbed past 350k by ~21m elapsed.</p></li><li><p>Claim/e2e p99 reached ~67s.</p></li><li><p>Completion stayed below offered rate.</p></li></ul><p>Diagnostics taken during that failure showed the hot claim call had become expensive under backlog. One 512-row <code>claim_ready_runtime</code> sample took ~166ms and ~57k shared-buffer hits.</p><p>The issue was not dead tuples. It was claim-side duplicate-claim fencing doing per-row <code>ready_claim_attempt_batches</code> probes even after the head row had already been proved live.</p><p>I changed the common live-head claim path so it filters only ready tombstones. The <code>ready_claim_attempt_batches</code> ledger is still used by the stale-cursor recovery path when the head row is already spent, preserving the lost post-commit claim-cursor-advance recovery semantics.</p><h2>Local validation</h2><ul><li><p><code>cargo fmt --check</code></p></li><li><p><code>git diff --check</code></p></li><li><p><code>cargo check -p awa-model</code></p></li><li><p><code>cargo test -p awa-model normalize_claim_cursor_advances -- --nocapture</code></p></li><li><p><code>./correctness/run-tlc.sh storage/AwaSegmentedStorage.tla</code> passed</p></li><li><p><code>AwaSegmentedStorageTraceLostClaimAdvance.cfg</code> reached the expected trace-completion violation, confirming the lost-advance trace is still accepted</p></li><li><p>PR CI is now all green</p></li></ul><h2>Short smoke after the fix</h2><p>Run: <code>custom-20260615T232405Z-7454ac</code></p><p>Configuration:</p><ul><li><p>5k/s offered</p></li><li><p>512 workers</p></li><li><p>6m clean phase</p></li></ul><p>Results:</p><ul><li><p>Median completion: 4,991/s</p></li><li><p>Mean completion: 5,001/s</p></li><li><p>Median depth: 126</p></li><li><p>Max depth: 40,010</p></li><li><p>Final depth: 125</p></li><li><p>Median claim p99: 150ms</p></li><li><p>Peak sampled table dead tuples: 126</p></li><li><p>Summary peak dead tuples: 423</p></li><li><p>No slow-statement, stale-heartbeat, or error warnings</p></li></ul><h2>Full local long-shape run after the fix</h2><p>Run: <code>custom-20260615T233249Z-eee140</code></p><p>Configuration:</p><ul><li><p>5k/s offered</p></li><li><p>512 workers</p></li><li><p>Warmup + 4 × 15m clean phases</p></li><li><p>Default settings</p></li></ul><p>Final depth returned to baseline at 126, and the warnings scan was clean.</p>
Phase | Median enqueue/s | Median complete/s | Mean complete/s | Median depth | Max depth | Final depth | Median claim p99 | Max claim p99 | Peak dead tuples | WAL delta
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
clean_15m_a | 4,991.7 | 4,991.3 | 4,974.9 | 126 | 37,541 | 20,767 | 938ms | 7.5s | 491 | 4.25GB
clean_15m_b | 4,991.7 | 4,990.9 | 5,021.3 | 1,312 | 30,343 | 3,429 | 2.27s | 7.0s | 565 | 4.18GB
clean_15m_c | 4,991.3 | 4,794.9 | 4,968.1 | 6,761 | 71,796 | 31,386 | 3.42s | 14.6s | 535 | 4.20GB
clean_15m_d | 4,991.5 | 4,990.0 | 5,035.6 | 3,132 | 63,244 | 126 | 3.00s | 13.1s | 578 | 4.29GB

<h2>Interpretation</h2><p>This closes the previous Phase-B runaway.</p><p>The remaining mechanism is now bounded claim/WAL/commit burst-and-catch-up behavior, not unbounded MVCC dead-tuple accumulation.</p><p>It can still create multi-second p99 tails and transient depth bursts, especially in later phases. I would describe the result as:</p><blockquote><p>This machine can absorb a sustained 5k/s offered single-queue workload with the current defaults, but not as a flat low-latency steady state.</p></blockquote><p>The next throughput/latency lever is still reducing periodic durable claim/commit stalls, or using first-class queue fanout where partitioned ordering is acceptable.</p></body></html>

Issue #295, opened 2026-06-01
RFC: v0.7 storage engine — segment storage + cursor allocator


hardbyte:
## Status

**Design RFC, not implementation.** This issue exists to collect input on the *shape* of the next storage engine before any code is written. Comments and pushback welcome.

## Motivation

The 0.6 release blocker work in #169 confirmed (long-horizon benchmark, 2026-05-17) that even with ADR-012/023/025/026 in place, Awa's queue_storage degrades under a pinned MVCC horizon — 799 → 387 jobs/s at 800 offered/s in a 2h pinned phase. That is the same class of behaviour every per-row-state-machine Postgres queue has (River, Oban, pg-boss, Graphile, pgmq); pgque holds at 799/s by trading feature surface (no per-row retries, heartbeats, cancellation) for immunity.

The 0.6 plan is to land the live terminal counter (#290) and an open-ended HOT-update audit pass to take the visible hot-table scans out of the operator path. That treats the symptoms. The architectural fix lives one level deeper: the lifecycle table itself is the wrong shape for sustained, long-reader-exposed workloads.

The spike investigation captured this in [`docs/issue-169-storage-spike.md`](../docs/issue-169-storage-spike.md). The high-level direction:

- **Append-only rotation segments** for job lifecycle, not a mutable lifecycle table.
- **Tiny per-segment claim ledger** (the only mutable hot rows in the design).
- **Cursor allocator** moving readers forward across segments rather than scanning prunable history.

The spike concluded the storage *shape* is right but the *claim allocator* is the load-bearing piece — the design has to start there, not at the segment format.

## Scope of this RFC

What we want input on:

1. **Claim allocator design.** What is the simplest allocator that gives SKIP LOCKED-equivalent fairness, supports per-row retries / heartbeats / cancellation, and does NOT degrade under a pinned MVCC horizon? Is it a single advisory-locked cursor table? A per-shard cursor with hash-routed jobs? Something else?
2. **Receipt plane integration.** ADR-023's ring partitioning + ADR-026's narrow terminal history are good ideas that should survive the redesign. How do they compose with rotation segments?
3. **Migration story.** Awa just did a 0.5→0.6 canonical→queue_storage migration. A 0.6→0.7 segment-storage migration is a real cost. What does the cutover shape look like — staged like 0.5.x→0.6, or a hard cutoff?
[...]

Issue #197, opened 2026-04-27
0.6 release readiness: align TLA+ models, docs, and queue-storage replacement gates


hardbyte:
## Context

Before `0.6` replaces canonical storage with queue storage, we need one explicit release-readiness tracker for the remaining correctness, documentation, and validation gates that sit above individual feature issues.

Existing milestone context:

- #180 tracks the rolling `canonical -> queue_storage` migration mechanics and operator flow.
- #169 tracks MVCC / dead-tuple behavior under sustained load.
- #174 tracks unifying the portable chaos and long-horizon runner.
- #167 tracks reducing chaos-test dependence on stdout/timing.
- #160 tracks a DLQ pagination correctness bug that matters if DLQ is part of the 0.6 operator story.
- #186 tracks Python packaging of `awa serve`.

This issue is not a duplicate of those. It is the release gate that says: queue storage is ready to replace canonical only once the formal model, docs, transition story, and long-horizon evidence line up.

## Current local state snapshot — 2026-04-28

Latest local branch head reviewed: `6e295cb` (`Merge origin/main (0.5.6) into 0.6 branch`).

Key deltas since the last issue update:

- Migration stack is now `CURRENT_VERSION = 13`:
  - `v011_storage_transition_self_heal.sql` — 0.5.x-series transition-state self-heal.
  - `v012_queue_storage_compat.sql` — queue-storage compat layer / active backend selection.
  - `v013_storage_auto_finalize.sql` — fresh-install auto-finalize plus queue-storage count maintenance.
- The second 12h ADR-023 overnight validation is recorded in `docs/adr/bench/023-receipt-ring-validation-2026-04-26.md`.
  - Receipt-plane invariant held: `lease_claim_closures_0..7` stayed at 0 dead tuples, `lease_claims_0..7` peaked at 24, RSS stayed flat, and the run had no ERROR/FATAL/panic.
  - The run exposed a separate throughput collapse under multi-million-row backlog, traced to exact `queue_counts` scanning `ready_entries`.
- The `queue_counts` bottleneck has been fixed in the local branch:
  - `queue_counts_exact` reads `sum(queue_lanes.available_count)` instead of scanning `ready_entries`.
  - dispatcher claimer sizing uses a dedicated `AvailableSignal` from the same counter.
  - `queue_counts_cached`, `queue_count_snapshots`, and `CLAIMER_QUEUE_COUNTS_MAX_AGE` are gone.
[...]

PR #355, opened 2026-06-18
Stop writing the queue_claim_heads routing cache; route via indexed segment scan

hardbyte:
## Problem

The v0.6 MVCC gate (60-minute idle-in-transaction soak @ 800 jobs/s) showed `queue_claim_heads` accumulating **130,199 dead tuples on a single live row** — 80.5% of all dead tuples in the run — because the `ready_segment_*` routing cache is `UPDATE`-d in place on nearly every claim, and a pinned snapshot blocks reclamation until the horizon releases. After the terminal-counter striping landed, this was the dominant remaining hot-row chain under a pin.

This is the `queue_claim_heads` finding from the HOT-update audit tracked in #197, and the bench re-run feeds the #169 "HOT-update audit pass complete with at least one bench re-run" acceptance item. It is **not** the v0.7 segment-engine work (that stays deferred); it's a direct continuation of the `queue_lanes.available_count` cache removal (v016, #251) and the `leases`/`lease_claims` fillfactor pass (#315).

## Investigation

An EXPLAIN study (50k-segment synthetic) showed the cache wasn't the right fix — the real cost was the cache-miss scan using `ORDER BY first_lane_seq` against a `next_lane_seq > cursor` predicate: not partition-prunable, and unable to short-circuit, so it **materialised + heapsorted the entire tail** (rows growing with the claim cursor). The cache was hiding a bad query plan — and was itself walking 130k dead tuples on its bloated row under the pin.

## Fix

- Remove the `ready_segment_*` cache; resolve the target ready slot from `ready_segments` (the existing compact "claim allocator" control plane) on every claim.
- Order the segment lookup by `next_lane_seq ASC LIMIT 1`. Ready segments are disjoint, contiguous lane ranges, so the smallest `next_lane_seq > claim_seq` is the same containing segment as the smallest `first_lane_seq`, but it lets the existing `(queue, priority, enqueue_shard, next_lane_seq, …)` index short-circuit (Merge Append, one index row per partition, `LIMIT 1` halts). ~0.17 ms warm and no longer grows with the cursor.
- No new index, no hint table, no new machinery. `queue_claim_heads` returns to the cold lane-registry shape #295 intended.

## Results — 60-minute pinned-MVCC gate, awa-only @ 800/s, PG 18.3

Compared against the previous cache (same harness, same shape):

| metric | cache (before) | this PR |
|---|---:|---:|
| `queue_claim_heads` dead tuples (idle peak) | 130,199 | **0** |
| total dead tuples (idle peak) | 161,944 | **31,178** (−81%) |
| `claim_p99` under pin | 88 ms | **55 ms** (−38%) |
| `claim_p99` steady-state | 24 ms | **22 ms** |
| e2e p95 under pin (median / max) | 38 / 17,859 ms | **23 / 1,266 ms** |
| queue depth max under pin | 14,265 | **450** |
| completion (all phases) | 798/s | **798/s** |

The cache-free path is faster under the pin than the cache it replaces, because the cache's own reads were walking the bloated singleton row.

**Known transient:** `claim_p99` is elevated in the post-pin recovery window (~165 ms vs the cache's ~42 ms) while autovacuum clears `ready_segments` bloat accumulated during the hour. It is not user-facing — recovery e2e p95 is 26 ms, completion holds 798/s, depth drains normally — and decays as vacuum runs. The lever, if ever needed, is autovacuum tuning on the `ready_segments` partitions, not the query.

## Migration / compatibility

- `v039` re-runs `install_queue_storage_substrate` per schema so the cache-free `claim_ready_runtime` reaches `awa` and any in-transition `awa_exp`. `CURRENT_VERSION` → 39.
- The now-unused `ready_segment_*` columns are **retained**, not dropped — dropping them violates the additive-only migration policy (`migrations.rs`) and would break a rolling worker whose `queue_storage_schema_ready` check still requires them. A future major version can drop them. (Updated from the first revision per review: the dead-tuple win comes from the function no longer *writing* the cache, not from removing the columns.)

## Docs

Brought the docs/model-mapping in line with the cache-free claim path: `docs/architecture.md`, `docs/queue-storage-substrate.md`, `docs/upgrade-0.5-to-0.6.md`, `correctness/storage/MAPPING.md`, and a `**historical:**` note in `docs/adr/019`. The claim comments + CHANGELOG now say ready segments are *non-overlapping* (aborted sequence reservations can leave gaps), so the `next_lane_seq` probe picks the next covering segment and the `ready_entries` validation prevents emitting through a gap.

## Verification

- `cargo fmt` · `clippy --all-targets --all-features -D warnings` · `build --workspace` — clean.
- `migration_test` (43) + `queue_storage_runtime_test` (112) — green. The migration test asserts the columns are retained; the obsolete cache-population runtime test is removed (the function no longer writes the cache).
- TLC `AwaSegmentedStorage` — no error (the claim cursor is a logical counter in the spec; the physical routing cache is below its abstraction, so the model is unchanged).

Peak-throughput ceiling deferred to nightly benchmarks; steady-state `claim_p99` parity (22 vs 24 ms) makes a ceiling regression unlikely.

Refs #169 #197

## Summary by CodeRabbit

* **Improvements**
  * Updated claim routing to resolve the target ready slot directly from live ready-segment data on each claim, removing reliance on the previous routing cache.
* **Chores**
  * Added schema migration to version 39 to enable the cache-free routing path, while retaining legacy cache columns for rolling-upgrade compatibility.
* **Tests**
  * Replaced the cache-population test with coverage that verifies the legacy cache columns are not written during claims.
* **Documentation**
  * Updated architecture and upgrade/migration guides to mark the legacy cache columns as now-unused but retained.


hardbyte:
Validation/review pass on #355.

I think the underlying mechanism and fix are sound. The per-claim `queue_claim_heads.ready_segment_*` routing cache was reintroducing a mutable hot singleton row under pinned MVCC; removing it and resolving from `ready_segments` on each claim matches queue-storage's append/sequence direction. The `next_lane_seq > claim_seq ORDER BY next_lane_seq LIMIT 1` probe looks valid under lane reservation ordering, with the later `ready_entries` validation preserving safety across sequence gaps or uncommitted reservations.

Local validation I ran:

- `cargo fmt --check`
- `cargo check -p awa-model`
- `cargo test -p awa-model ready_segments_split_on_run_at_boundaries -- --nocapture`
- With temporary Postgres 18 on `localhost:15432`:
  - `test_insert_job_compat_routes_under_active_queue_storage_engine`
  - `test_migration_sql_range_produces_valid_schema`
  - `test_queue_storage_claim_gap_does_not_skip_uncommitted_enqueue_sequence`
- TLC:
  - `storage/AwaDeadTupleContract.tla`
  - `storage/AwaSegmentedStorage.tla`
  - `storage/AwaStorageLockOrder.tla`

GitHub CI was all green when checked.

Findings I would clean up before merge/release:

1. Some docs/model mapping references still describe the removed `queue_claim_heads.ready_segment_*` cache. The affected places I found were `docs/architecture.md`, `docs/upgrade-0.5-to-0.6.md`, `docs/queue-storage-substrate.md`, `docs/adr/019-queue-storage-redesign.md`, and `correctness/storage/MAPPING.md`. These now contradict v039 and `queue_storage_schema_ready()`.

2. The SQL comment in `v023_install_queue_storage_substrate.sql` says ready segments are "disjoint contiguous lane ranges". The implementation is still safe, but aborted sequence reservations can leave gaps, so I would tighten that to something like "non-overlapping and visibility-ordered" and note that `ready_entries` validation prevents emitting through gaps.

No correctness blocker found in the SQL/migration path. This looks like a high-value removal of the claim-head routing-cache MVCC pile-up once the stale docs/mapping text is brought in line.

PR #482, opened 2026-09-06
Add owner-scoped periodic reconciliation and durable retirement

### LINKED ISSUE #481: Owner-scoped declarative reconciliation for periodic schedules
## Problem

AWA intentionally synchronizes periodic schedules additively. Removing a
schedule from worker code leaves its row active until an operator runs
`awa cron remove <name>` or deletes it with SQL. ADR-007 records this as the
safe trade-off because multiple independent deployments may share the same
AWA database; a process cannot assume that every schedule it did not declare
is an orphan.

This is safe, but it is easy for a code removal and the corresponding manual
cleanup to drift apart. A stale schedule can continue enqueueing work with no
runtime that intentionally owns it.

We should support **opt-in, owner-scoped declarative reconciliation** without
introducing global pruning.

## Research / prior art

- [Oban DynamicCron](https://oban.hexdocs.pm/2.11.0/dynamic_cron.html)
  supports an explicit `delete: true` declaration. This is a useful minimal
  DX improvement because deletion is named and intentional rather than inferred
  from global absence.
- [Kubernetes Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/)
  records manager identity and rejects conflicting ownership rather than
  silently overwriting another manager.
- [Kubernetes ApplySet pruning](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#applyset-based-pruning)
  tracks set membership before pruning. This is the relevant model for
  reconciling a complete owned collection; a label/owner alone is not enough.
- [River periodic jobs](https://github.com/riverqueue/river/blob/master/periodic_job.go)
  require additions/removals to reach every active client because only the
  elected leader acts on its local configuration. AWA has the analogous
  topology today: only the maintenance leader periodically upserts its local
  `periodic_jobs` vector.

## Safety observations from the spike

1. An owner ID alone does not establish that a process has declared the
   owner's **complete** desired set. A client may declare only a subset, and an
   accidentally empty vector must not mean "remove all".
2. During a rolling deployment, old and new replicas legitimately declare
   different sets. Reconciliation must not oscillate as leadership moves.
3. Current schedule sync is leader-local. Reconciliation cannot infer fleet
   convergence from the leader's vector; participating instances must publish
   desired-set evidence independently of leadership.
4. `paused_at` is not the right lifecycle state. Resuming a paused `catch_up`
   schedule may replay missed fires, while a schedule removed from desired state
   must stay inert until explicitly restored.
5. Deleting a row is not a sufficient old-runtime fence. An old leader may
   recreate it through the existing additive UPSERT. A durable retired row can
   survive that UPSERT, while a database-side guard can make the old atomic
   enqueue CTE affect zero rows.
6. A vanished fleet is indistinguishable from an outage. No schedules should be
   retired merely because all instances for an owner stopped heartbeating.

### 1. Preserve additive sync as the default

Unowned schedules and clients that do not opt in retain today's ADR-007
behaviour. There is no global `DELETE ... NOT IN (...)` path.

Add a stable, non-empty reconciliation owner/manager ID to client configuration,
for example (names illustrative):

```rust
ClientBuilder::new(...)
    .periodic_owner("billing-workers")
    .periodic_reconciliation(PeriodicReconciliation::Authoritative)
    .periodic(daily_invoice)
    .periodic(payment_reminders);
```

Python should expose the same semantics. `Authoritative` means "this is the
complete desired set for this owner", not merely "these schedules belong to
this owner". An authoritative empty set must be explicitly representable and
must not be inferred from a default/omitted configuration.

Schedule names should remain globally unique in the first iteration so existing
CLI/API/job metadata lookups by `cron_name` stay unambiguous. Ownership scopes
reconciliation; it does not initially change schedule identity to
`(owner, name)`.

### 2. Reject ownership conflicts

New schedules record `owner_id`. Updating a schedule owned by another owner
fails with a clear conflict. Claiming an existing unowned schedule or
transferring ownership is an explicit operation with a dry-run/confirmation
surface; registration must not silently adopt it.

An older additive UPSERT that does not name the new ownership columns must not
erase ownership or retirement state.

### 3. Publish desired-set evidence from every participating runtime

Each authoritative runtime publishes, independently of maintenance leadership:

- runtime instance ID;
- owner ID;
- deployment revision (opaque, for diagnostics/grouping);
- canonical desired-set hash;
- desired schedule names or an equivalent normalized manifest;
- last-seen time / snapshot interval.

This can extend the existing runtime snapshot/reporting path or use a separate
declaration table keyed by `(instance_id, owner_id)`.

The maintenance leader may reconcile an owner only when:

- every fresh runtime in the shared AWA database reports support for the
  ownership/declaration protocol (older runtimes cannot be safely attributed to
  an owner and must conservatively block retirement);
- at least one fresh authoritative declaration exists;
- every fresh declaration for that owner has the same desired-set hash;
- that agreement has remained stable for a configurable grace period;
- the transaction re-checks the same conditions while holding a per-owner
  advisory/row lock, closing the check-to-retire race.

Mixed desired-set hashes block retirement and surface an observable reason.
Additions/definition updates may continue safely while retirement is blocked.

Do not retire anything when there are zero fresh declarations. Decommissioning
an entire owner requires an explicit `awa cron retire-owner <owner>` (or
equivalent API/UI action).

### 4. Retire absent schedules; do not immediately delete them

Add a lifecycle state such as `retired_at`, `retired_by`, and
`retired_revision` (exact schema TBD). Reconciliation atomically marks owned
schedules absent from the converged desired set as retired.

The evaluator must skip retired rows and `atomic_enqueue` must re-check that the
row is not retired in the same UPDATE/CAS that advances `last_enqueued_at`.
Add a database-side fence (for example, a trigger or compatible function
boundary) so an old binary's existing enqueue CTE cannot advance/fire a retired
row. This must be tested using the actual released N-1 binary path, not a
sentinel query.

Re-declaring a retired name should not silently resurrect it. Restoration is an
explicit action so a stale binary/config cannot undo retirement.

Initial scope should retain retired rows indefinitely. A later purge policy can
be considered separately after its mixed-version and audit implications are
specified.

### 5. Operator and observability UX

Provide a plan/dry-run surface that reports:

- additions and updates;
- schedules that would be retired;
- ownership conflicts;
- mixed/non-converged runtime declarations;
- zero-live-declaration refusal;
- the revision/hash and grace-window evidence used for the decision.

Expose owner, active/retired state, reconciliation status, and blocking
instances in the CLI, admin API, UI, logs, and metrics. Manual per-schedule
remove/retire remains available.

## Migration and rollout requirements

This must follow ADR-041 and the migration checklist:

- expand additively; do not change or drop the existing primary key initially;
- preserve operation of the released N-1 binary on the expanded schema;
- if the database-side retirement fence requires a compatibility patch, ship
  and verify that patch first and add the corresponding migration runtime floor;
- document requirements for external SQL migration runners;
- rehearse migrate-first, binary-first, overlapping old/new leaders, leadership
  transfer, stale/reappearing runtimes, crashes during reconciliation, and an
  explicitly authoritative empty set;
- update ADR-007, CHANGELOG, upgrade documentation if operator action is
  required, `docs/stability.md` if the skew contract changes, and the
  `awa-jobs` / `awa-operations` skills as appropriate.

The reconciliation transaction and retirement/enqueue interaction should be
added to the cron TLA+ model.

## Acceptance criteria

- [ ] Default schedule sync remains additive and backward compatible.
- [ ] Two owners sharing one database cannot update or retire each other's
      schedules.
- [ ] One authoritative client cannot retire schedules from an unrelated
      additive client.
- [ ] A partial declaration or accidental empty/default registration cannot
      trigger retirement.
- [ ] Differing live desired-set hashes during a rolling deploy block
      retirement; convergence for the grace period permits it.
- [ ] A fresh runtime that does not report ownership/reconciliation capability
      blocks retirement, even if all visible declarations for an owner agree.
- [ ] Zero live declarations never trigger automatic retirement.
- [ ] An explicit authoritative empty set can retire all schedules for exactly
      one owner after the convergence gate.
- [ ] The reconciliation decision is transactionally re-checked so a fresh
      conflicting declaration cannot race a retirement commit.
- [ ] Retired schedules cannot enqueue through current or released N-1 atomic
      enqueue paths, including after leadership transfers back to an old binary.
- [ ] Old additive UPSERTs cannot clear ownership or retirement state.
- [ ] Re-declaring a retired schedule does not resurrect it without an explicit
      restore/adopt action.
- [ ] Dry-run output explains every proposed change and every safety refusal.
- [ ] Rust and Python APIs have equivalent ownership/reconciliation semantics.
- [ ] CLI/API/UI and documentation distinguish paused from retired schedules.

## Non-goals

- Global pruning of schedules absent from one process.
- Automatically deleting an owner's schedules because its whole fleet is down.
- Cancelling already-enqueued, running, retrying, or DLQ jobs when their source
  schedule is retired.
- Changing schedule identity to a composite `(owner, name)` key in the first
  iteration.
- Automatic physical purge of retired rows.

hardbyte:
Removing a periodic schedule from code leaves it running today. This adds opt-in owner-scoped complete manifests: the fleet must agree through a durable grace window before absent schedules retire. Omitted configuration remains additive; explicit empty manifests are supported. Schema v045 preserves ownership and retirement tombstones through released 0.6.7 UPSERTs and fences its automatic cron enqueue. Existing jobs continue independently.

Rust/Python and CLI/API/UI expose plans, adoption/transfer, retirement and restoration. Owned-schedule operator actions require current clients because old manual triggering uses a separate insert path. Restoration preserves pause state and starts from current database time.

## Review disposition

The [September 6 review](https://github.com/hardbyte/awa/pull/482#issuecomment-5557812027) is addressed in `b725b27`, `c0f00e6` and `59a1377`:

- Existing definitions wait for fresh capable fleet agreement; mixed manifests no longer overwrite each other. New names remain additive. Definition synchronization is independent of retirement grace.
- Lifecycle actions lock only selected rows and invalidate agreement only for affected owners, including both sides of a transfer.
- Retired unowned additive registration is a quiet no-op. Physical deletion reports the retained tombstone accurately. Same-owner retired desired names allow inert startup; `restore-owner` restores the owner's retired schedules in bulk.
- Plans, previews and dry runs use a consistent read-only snapshot without the global writer lock. Apply rechecks under that lock. Already-cleared legacy evidence is not rewritten.
- Heartbeats cannot certify a conflicting manifest; unchanged synchronized manifests avoid repeated deserialization.
- Protocol version is named and checked against SQL, migration constants are ordered, metrics have descriptions, no-op reconciliation logs at debug, and UI audit labels default to the serving instance (not authenticated user identity). The vacuous TLA stutter was removed. The preview test took 2.12s in isolation; no timing bound was relaxed.

Final-head rehearsal also exposed a real migration lock-order cycle: pending storage DDL preceded v045's cron lock while released enqueue took cron first. The runner now drains cron before the entire pending DDL range. A deterministic SQL regression observed 50 conflicting DDL locks before the fix and zero after; a small TLA model preserves the missing-drain counterexample. Cron evaluation pauses until migration commit; the upgrade guide and operations skill specify the same ordering for external SQL runners.

## Validation and release status

- Required Rust formatting, all-target/all-feature clippy and workspace build pass. Five review regressions were demonstrated red before fixes. The final focused cron suite passes 28 tests (one released-artifact test is separately gated).
- Released 0.6.7 live cron rehearsal passes locally with the lock fix on PG17 and PG18. Models cover ownership safety/liveness, mixed-definition gating and migration lock order, including expected negative witnesses.
- Review-round Python periodic tests: 8 passed; API tests: 7 passed; browser tests: 10 passed; frontend build and both skills validate.
- [Full CI: 29/29 jobs](https://github.com/hardbyte/awa/actions/runs/34026929816) and [release gate: all three jobs](https://github.com/hardbyte/awa/actions/runs/34026762355) passed on exact head `59a1377`. Released-artifact rehearsals, protocol models and expected negative witnesses pass. The interrupted local full-suite run is not counted as a pass.
- The pinned old-wheel teardown crash remains unresolved: [0.6.2 exited 139 after lifecycle PASS](https://github.com/hardbyte/awa/actions/runs/34025905890/job/101466701977), despite explicit pool cleanup. The earlier 0.6.0 failure remains evidence too. A green retry does not establish a native root-cause fix.

## Performance

The [fresh overnight campaign](https://github.com/hardbyte/postgresql-job-queue-benchmarking/blob/brian/awa-481-benchmarks/results/2026-09-06-awa-482-overnight/SUMMARY.md) completed in 8h 44m, with main `49b1a77` versus candidate `59a1377`: all ten throughput/soak cells, nine cron probe configurations, and both four-hour soaks passed. Each two-hour MVCC pin validated all 1,440 samples and release. Both builds sustained ~800 completions/s with median queue depth zero; pinned median window-p99 was 44ms / 46ms. Saturation throughput deltas were +6.3% (W64), −3.7% (W128), −3.3% (W256), each one sequential pair.

Candidate first crossed ≤10% of pinned dead-tuple peak after 45s and ≤110% of its clean median after 19m 45s; main crossed at the first sample and after 35s. Neither build stayed below its clean threshold throughout recovery. Tracked relations stayed below 25 MiB. The [interpretation and consistent plots](https://github.com/hardbyte/postgresql-job-queue-benchmarking/blob/brian/awa-481-benchmarks/results/2026-09-06-awa-482-overnight/INTERPRETATION.md) retain that recovery difference and the limits of single-pair inference. The Aug 23 soak is discarded.

Both builds use the same pinned [upstream SQLx TCP_NODELAY fix](https://github.com/transact-rs/sqlx/pull/4336). AWA keeps SQLx 0.9; these results do not validate the unpatched published dependency. A fixed published SQLx release and shipping-dependency validation remain release blockers.

Also updates dependencies, including seroval 1.6.4 (superseding closed #453).

Fixes #481.

## Summary by CodeRabbit

* **New Features**
  * Added owner-scoped cron reconciliation with complete schedule manifests, status planning, adoption, retirement, restoration, and ownership transfer.
  * Added CLI and Python client support for previewing and applying cron lifecycle actions.
  * The cron UI now displays ownership and retirement status, reconciliation blockers, and preview/apply controls.
  * Retired schedules are durably inactive and cannot be triggered or automatically enqueued.
* **Bug Fixes**
  * Improved migration and enqueue coordination to prevent lock-related conflicts.
* **Documentation**
  * Added configuration, upgrade, stability, and operational guidance for cron ownership and reconciliation.


hardbyte:
## Review of 4c2fa83

Verified locally on PG17: `cargo fmt --check`, all-feature clippy with `-D warnings`, `cron_reconciliation_test` (20 passed, released-probe test ignored), and `awa-ui` `cron_api_test` (7 passed). Not run here: Python suite, Playwright, `rehearse-cron-ownership.sh`, TLC. Note that full CI has only run at `1bddb2e`; only the release gate ran at the current head, so re-run full CI on the final head before merging.

The migration, fence trigger, lock ordering, and the startup publish/check ordering all read correctly, and the acceptance criteria in #481 are covered by tests. The items below are things I confirmed with throwaway tests against this branch, followed by design questions and nits.

### Should fix before merge

**1. Definition flip-flop during a mixed-manifest rolling deploy** (`awa-model/src/cron_reconciliation.rs`, `publish`)

`synced_hash` is a single value per owner. With instance A on rev1 (`x` = `0 9 * * *`) and B on rev2 (`x` = `0 10 * * *`), each heartbeat sees `synced_hash != hash` and re-runs the batch UPSERT, so the stored `cron_expr` alternates at heartbeat cadence for the whole deploy:

```
after_b=0 10 * * *  after_a=0 9 * * *  after_b2=0 10 * * *
```

Before this PR the definition was leader-local and stable for the duration of a leader term. Retirement is already gated on a single live hash; definition sync should be gated the same way (or track synced state per `(owner, hash)` and only write when the live set agrees). New names could still be inserted immediately since that is harmless. Worth a test alongside `transient_conflict_resets_grace_between_leader_passes`.

**2. Operator actions reset every owner, and lock every schedule row** (`operate`)

```rust
sqlx::query("UPDATE awa.cron_owners SET synced_hash=NULL,agreed_since=NULL,evidence_until=NULL")
```

has no `WHERE`, and the apply path does `SELECT * FROM awa.cron_jobs ORDER BY name FOR UPDATE` over the whole table. Retiring one `billing` schedule reset `mail`'s `agreed_since` and `synced_hash` in my probe, which restarts an unrelated owner's grace window and forces its instances to re-run the batch UPSERT. #481's isolation goal argues for scoping both to the affected owner(s) and the selected rows.

**3. Retiring an unowned schedule turns additive clients into a permanent error stream** (`cron::upsert_cron_job`, `awa.cron_schedule_fence`)

`Retire { name }` accepts unowned rows. Any additive deployment still declaring that name then fails `upsert_cron_job` every cron-sync pass and logs at `error!` ("Failed to sync periodic job"). That is the intended tombstone behaviour, so it should be a quiet skip (debug, or a single warn) rather than an error, and distinguishable from the genuinely wrong case of colliding with another owner's schedule. Also, `awa cron remove` on that row raises "has durable ownership; retire it instead of deleting", which is misleading for a row that has no owner and is already retired.

### Design questions

**4. Startup failure on a retired desired name.** `check_registration` fails `start()` if a declared name is retired. The common path to that state is: remove schedule, deploy (retired), later re-add the schedule, deploy. The second deploy crash-loops until an operator runs `awa cron restore` mid-rollout. After `retire-owner` every schedule is in that state and there is no bulk restore. Failing startup on a foreign-owner conflict is defensible; for retired names I'd rather surface it as the existing `ownership_or_retirement_conflict` blocker (the schedule stays inert, which is the safe outcome) and add a `restore-owner`/`--all` operation. If you keep the current behaviour, the upgrade guide should call out the crash-loop explicitly.

**5. Control-plane lock scope.** Every `runtime_instances` write in the fleet now serialises on one `pg_advisory_xact_lock`, and read-only `plan`/`preview` take it too. The UI polls `GET /api/cron/reconciliation` (one `plan` per owner) at the page poll interval, so a dashboard left open contends with heartbeats. Probably fine at realistic fleet sizes, but read-only plans could skip the lock. Separately, `cron_unknown_runtime` issues an unconditional `UPDATE awa.cron_owners` on every legacy heartbeat even when the columns are already NULL; a `WHERE agreed_since IS NOT NULL OR evidence_until IS NOT NULL` avoids the dead tuples.

**6. Agreement jitter under conflicts.** The heartbeat publish path (`persist && !apply`) skips `describe_changes`, so conflicts don't block persisting agreement there, while the leader pass does see them and resets it. Result: `agreed_since`/`grace_remaining_ms` oscillate between heartbeat and leader passes whenever a conflict exists. Harmless for safety, but confusing in `awa cron plan` output.

### Nits

- `admin.rs` hardcodes `cron_protocol = 1`; a named const referenced from the INSERT and the test helper would keep it in step with `awa.cron_protocol_version()`.
- `V45_UP` is declared before `V44_UP` in `migrations.rs`.
- `LegacyUpsert == UNCHANGED vars` in `AwaCronOwnership.tla` is a stutter step and models nothing. Deletion isn't in the model, and that is what the tombstone defends against; either model delete plus legacy recreate or drop the action.
- The two new metrics lack `.with_description(...)`, unlike the neighbouring counters.
- The UI posts `actor: "web-ui"`; the router already has identity via `router_with_identity`, which would make the audit column useful.
- `sync_periodic_jobs_to_db` logs `info!` per owner per pass forever, including fully decommissioned owners; consider `debug!` when nothing changed.
- `proposed_manifest_is_read_only_and_explains_all_changes` took over 60s locally while its siblings took under a second; worth a look for CI flakiness.

